PR #97, opened 2014-11-28
Spike/view data

matharden:
Display the prison data in an human readable way.


matharden:
Closing until we decide how we want to display this data publicly.

PR #123, opened 2014-12-30
Added a hidden link to the current commit to the body of the page.

jasiek:
Please don't merge until @mikepea has had a chance to ensure that the META file in question is, in fact JSON.


jasiek:
I really wanted to avoid having to view the source.

PR #124, opened 2014-12-30
Closed visits 84122476

matharden:
Add a checkbox to approved visit requests to notify visitors when the visit will be a 'closed visit'.

@jasiek please look at the [first commit](https://github.com/ministryofjustice/prison-visits/commit/17067293fa5c0ce96e73f9ae8726bd4fb56734be) first as I'm not convinced I've implemented the best way to replicate the token.

### DIFF
--- a/app/views/deferred/confirmations/_canned_responses.html.haml
+++ b/app/views/deferred/confirmations/_canned_responses.html.haml
@@ -41,11 +41,10 @@
         = text_field_tag 'confirmation[vo_number]', nil, :maxlength => 8
       
       .group
-        #closed_visit.field{ :class => ('validation-error' if f.object.errors.include?(:closed_visit)) }
+        #closed_visit.field
           %label.block-label(for='confirmation_closed_visit')
             = check_box_tag 'confirmation[closed_visit]', 1, f.object.closed_visit.present?, { :'data-conditional-el' => 'details_closed_visit', :class => 'js-Conditional' }
             This is a closed visit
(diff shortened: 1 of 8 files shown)


nicolekobilansky:
Agreed with @SteveMarshall 's point - needs a bit more explanation. If @leighmoney can make it sound snappy in one sentence then may be we don't even have to have a link to an explanation elsewhere...

matharden:
Rebase and ready to go! :rocket:

app/views/deferred/confirmations/_canned_responses.html.haml
```diff
@@ -39,6 +39,17 @@
...
+      .group
+        #closed_visit.field
+          %label.block-label(for='confirmation_closed_visit')
+            = check_box_tag 'confirmation[closed_visit]', 1, f.object.closed_visit.present?, { :'data-conditional-el' => 'details_closed_visit', :class => 'js-Conditional' }
+            This is a closed visit
+
+        #details_closed_visit
+          %h3 What we’ll tell the visitor
+          %p
+            %em.email This is a closed visit.
```
SteveMarshall: Does this need to exactly reflect the text we show the visitor, or is this enough?

matharden: Ah, yes - good spot. Thanks.

matharden:
:boom:

SteveMarshall:
:shipit: :rocket:

PR #128, opened 2015-01-02
Jasmine specs

matharden:


spec/javascripts/checkbox-summary_Spec.js
```diff
@@ -0,0 +1,72 @@
...
+      $fixture = $('<div class="CheckboxSummary"><input type="checkbox" value="Blue" id="Blue" /><span class="CheckboxSummary-summary">something else</span></div>');
+      $fixture.find('#Blue').prop('checked', true).change();
+      $fixture.find('#Blue').prop('checked', false).change();
+      expect($fixture.find('.CheckboxSummary-summary').text()).toBe('something else');
+    });
+  });
+
+  it('elements are cached', function() {
+    expect(subject.$checkboxes.length).toBe(2);
+  });
```
SteveMarshall: This test feels a bit like it's testing the module's internals, rather than testing the interfaces it presents to the world? Is there a way we can test the expected behaviour of the caching rather than inspecting the cache directly?

spec/javascripts/checkbox-summary_Spec.js
```diff
@@ -0,0 +1,72 @@
...
+
+  it('render is triggered from moj.Events module');
```
SteveMarshall: Should there be more to this test? It looks somewhat incomplete…

spec/javascripts/checkbox-summary_Spec.js
```diff
@@ -0,0 +1,72 @@
...
+      $fixture.find('#Orange').prop('checked', false).change();
+      expect($fixture.find('.CheckboxSummary-summary').text()).toBe('[summary]');
+    });
+
+    it('optional placeholder text is used when specified', function() {
+      $fixture = $('<div class="CheckboxSummary"><input type="checkbox" value="Blue" id="Blue" /><span class="CheckboxSummary-summary">something else</span></div>');
+      $fixture.find('#Blue').prop('checked', true).change();
+      $fixture.find('#Blue').prop('checked', false).change();
+      expect($fixture.find('.CheckboxSummary-summary').text()).toBe('something else');
+    });
```
SteveMarshall: It'd be good to have another test checking that `it('correctly summarises checked selections between every change')`. For example, make sure that it summarises two checkboxes (per `summary items are separated by commas by default`), then uncheck one of them and ensure that the summary is still correct.

SteveMarshall:
All your test titles read a little weirdly given the `it` method. Maybe tweak them (in a rebase and force push?) so they read like a sentence?

For example:

```js
it('original placeholder text is used when summary is blank'
```

might become:

```js
it('uses original placeholder text when summary is blank'
```

spec/javascripts/checkbox-summary_Spec.js
matharden: Yes, agreed. I'll add that one.

spec/javascripts/checkbox-summary_Spec.js
```diff
@@ -0,0 +1,95 @@
...
+    $fixture.find('#Orange, #Green').prop('checked', true).change();
+    expect($fixture.find('.CheckboxSummary-summary').text()).toBe('Orange, Green');
+    $fixture.find('#Orange').prop('checked', false).change();
+    expect($fixture.find('.CheckboxSummary-summary').text()).toBe('Green');
+  });
+
+  it('uses checkbox values as summary text', function() {
+    $fixture.find('#Orange').prop('checked', true).change();
+    expect($fixture.find('.CheckboxSummary-summary').text()).toBe('Orange');
+  });
```
SteveMarshall: Isn't this test a subset of `it('creates a summary from currently checked boxes'`? Can it be removed and just keep that one?

app/assets/javascripts/modules/moj.checkbox-summary.js
```diff
@@ -10,14 +12,20 @@
...
+        this.settings.original = sum;
+      }
+    },
+
+    defaults: {
+      glue: ', ',
+      strip: ';',
+      sub: ' ',
+      original: '[summary]'
     },
```
SteveMarshall: Can we come up with some better defaults for this? Replacing `;` with ` ` by default seems a bit weird to me, and it might be nice to have a nicer value for `original`.

spec/javascripts/checkbox-summary_Spec.js
matharden: It could, but I like how this documents the modules behaviour.

app/assets/javascripts/modules/moj.checkbox-summary.js
matharden: Yes, agreed.

spec/javascripts/checkbox-summary_Spec.js
matharden: Oh. It's the same as the one above the previous. Will remove.

SteveMarshall:
:shipit:

PR #137, opened 2015-01-09
Some prisons require that visits be booked at most two weeks in advance.

jasiek:
[#74286640]


jasiek:
Don't merge this yet please.

PR #138, opened 2015-01-09
Add temporary eCredits survey to request send page

matharden:


jasiek:
Looks good, provided this is temporary.

app/views/shared/_survey.html.haml
```diff
@@ -0,0 +1,9 @@
+#user-satisfaction-survey-container
+  %section#user-satisfaction-survey
+    .wrapper
+      %h1 Help us design new ways of sending money to prisoners
+      %p.right
+        %a#survey-no-thanks{:href => "#survey-no-thanks"} No thanks
```
SteveMarshall: Should this be a `button` rather than an `anchor`?

SteveMarshall:
I'm happy to :ship: this, but we need to check with @nicolekobilansky that the survey is ready to go…

SteveMarshall:
:eggplant:!

PR #141, opened 2015-01-19
Visit cancellation receipt contains VO number for reference.

jasiek:
[#86310816]

### DIFF
--- a/app/views/prison_mailer/booking_cancellation_receipt_email.text.erb
+++ b/app/views/prison_mailer/booking_cancellation_receipt_email.text.erb
@@ -2,6 +2,10 @@ Dear staff,
 
 <%= @visit.visitors.first.full_name %> has cancelled the visit to <%= @visit.prisoner.full_name %> (<%= @visit.prisoner.number %>) on <%= visit.slots.first.date %> <%= visit.slots.first.times %>.
 
+<% if @visit.vo_number %>
+Reference number: <%= @visit.vo_number %>
+
+<% end %>
 Please delete this visit from NOMIS.
 
 Kind regards,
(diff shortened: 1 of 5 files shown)


app/views/prison_mailer/booking_cancellation_receipt_email.text.erb
```diff
@@ -2,6 +2,10 @@ Dear staff,
 
 <%= @visit.visitors.first.full_name %> has cancelled the visit to <%= @visit.prisoner.full_name %> (<%= @visit.prisoner.number %>) on <%= visit.slots.first.date %> <%= visit.slots.first.times %>.
 
+<% if @visit.vo_number %>
+Reference number: <%= @visit.vo_number %>
+<% end %>
+
```
matharden: Please could you move this space inside the `if` block so it doesn't create a double space when there is no reference number?

spec/mailers/prison_mailer_spec.rb
```diff
@@ -94,12 +94,14 @@
     end
 
     it "sends an cancellation notification to a prison so it can be removed from NOMIS" do
+      sample_visit.vo_number = '87654321'
```
matharden: Might be good if this was moved into the :sample_visit so the VO number appears in mailer previews.

jasiek: Yeah, good idea.

app/views/prison_mailer/booking_cancellation_receipt_email.text.erb
```diff
@@ -2,6 +2,9 @@ Dear staff,
 
 <%= @visit.visitors.first.full_name %> has cancelled the visit to <%= @visit.prisoner.full_name %> (<%= @visit.prisoner.number %>) on <%= visit.slots.first.date %> <%= visit.slots.first.times %>.
 
+<% if @visit.vo_number %>
+Reference number: <%= @visit.vo_number %>
+<% end %>
```
matharden: Needs a line break here

PR #139, opened 2015-01-14
Fix: feature specs fail on date picker

matharden:
- regroup before blocks to allow email validation to pass and fail when required
- only allow instant bookings to find one slot per day
- deferred bookings slot times needed updating

Fixes 86118980

PR #145, opened 2015-01-20
Fix additional visitors blocking DOB after 2014

matharden:
2014 was hard-coded into the DOB year for additional visitors year using the `max` attribute. This commit replaces that with a rolling year. And also sets the `min` attribute to a rolling 115 years back.


SteveMarshall:
Does this need a test, @matharden? (To ensure we don't get a regression if this gets changed in future)

Also: is this being validated server side (I couldn't see anywhere that it was…)? Should it be?

matharden:
Yes there are server side tests to this effect.

matharden:
Yes, it would be good to have a test, but currently the feature tests are broken. I think #139 will fix them. So in the interest of allowing visits, I'd like to release this and add a test retrospectively.

SteveMarshall:
:+1:

PR #148, opened 2015-01-23
Update digital comms with latest two features

matharden:
- info on closed visits
- info on parent permission letters
- styling to support blockquotes written in Markdown


SteveMarshall:
:shipit:

PR #156, opened 2015-02-06
Per Mike's request, the initial redirect from "/" is configurable using ...

jasiek:
...an environment variable

(GOVUK_START_PAGE)


SteveMarshall:
:shipit:

PR #159, opened 2015-02-10
Nice, fortnightly performance metrics.

jasiek:
[#86694360]

### DIFF
--- a/app/assets/javascripts/metrics.js
+++ b/app/assets/javascripts/metrics.js
@@ -4,6 +4,9 @@
 
 'use strict';
 
+var  margin = {top: 10, right: 100, bottom: 100, left: 100};
+
+
 function percentile(array, n) {
     array.sort(function(a, b) {
         return a - b;
@@ -41,10 +44,13 @@ function formatSeconds(s) {
     return output.join(' ');
 }
 
+function formatDate(date) {
+    var months = ['January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December'];
+    return [date.getDate(), months[date.getMonth()]].join(' ');
+}
 function displayHistogram(where, dataSource, displayLines) {
-    var margin, width, height, x, y, n, data, maxY, svg, bars, xAxis, medianValue, percentileValue;
+    var width, height, x, y, n, data, maxY, svg, bars, xAxis, medianValue, percentileValue;
 
-    margin = {top: 10, right: 10, bottom: 30, left: 10};
     width = 960 - margin.left - margin.right;
     height = 500 - margin.top - margin.bottom;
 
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/metrics.js
```diff
@@ -156,3 +162,70 @@ function displayWeeklyBreakdown(where, rawDataSource) {
     
     chart.draw(processedData);
 }
+
+function thisYear() {
+    return parseInt(d3.select("#year").html());
+}
+
+function displayPerformanceLineChart(where, performanceData, volumeData) {
+    var xRange = [new Date(thisYear() - 1, 11, 29), new Date(thisYear() + 1, 0, 1)];
```
matharden: 29th Dec?

jasiek: Yeah...

According to ISO8601 the first week of the year can technically start in the previous year:

"It has 4 January in it. Hence the earliest possible dates are 29 December through 4 January, the latest 4 through 10 January."

app/assets/javascripts/metrics.js
```diff
@@ -41,10 +44,13 @@ function formatSeconds(s) {
     return output.join(' ');
 }
 
+function formatDate(date) {
+    var months = ["January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December"];
+    return [months[date.getMonth()], date.getDate()].join(" ");
```
matharden: I think we should display this date as date first. Ie "6 February".

lib/fortnightly_performance_report.rb
```diff
@@ -0,0 +1,35 @@
...
+FROM percentiles
+WHERE cume_dist >= 0.95
+GROUP BY fortnight
+ORDER BY x})
+  end
+
+  def volume
+    @model.connection.execute(%Q{
+SELECT MIN(DATE_TRUNC('week', requested_at))::date AS x, COUNT(*) AS y FROM visit_metrics_entries
+WHERE prison_name = '#{@prison_name}' AND EXTRACT(isoyear FROM requested_at) = #{@year}
```
SteveMarshall: Can we replace the string interpolations with parameters? I know this isn't connected to anything user-facing directly, but it might still leave us vulnerable to SQL injection… Something like:

```ruby
query = connection.raw_connection.prepare(%Q{
SELECT …
WHERE prison_name=? AND EXTRACT(isoyear FROM requested_at) = ?
…
})
query.execute(@prison_name, @year)
```

That way, ActiveRecord sanitises and cleans all the input for us so the variables can't be used to escape the statement and trash our DB.

app/assets/javascripts/metrics.js
```diff
@@ -41,10 +44,13 @@ function formatSeconds(s) {
     return output.join(' ');
 }
 
+function formatDate(date) {
+    var months = ["January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December"];
```
matharden: You're not going to like this, but can you stick with single quotes through this file please? So that JS Lint passes :see_no_evil:

app/assets/javascripts/metrics.js
jasiek: Yep.

lib/fortnightly_performance_report.rb
jasiek: Yep, will do.

app/assets/javascripts/metrics.js
matharden: I see. Is there some way you can make this more obvious in the code, so that somebody doesn't try to correct it to '31' in the future?

matharden: In that case does the end of the range not need to be 10 Jan also?

lib/fortnightly_performance_report.rb
matharden: :+1:

app/assets/javascripts/metrics.js
jasiek: I've no preference, will change it.

PR #161, opened 2015-02-12
Rewrote all queries as model.find_by_sql with bound params.

jasiek:


lib/detailed_metrics.rb
```diff
@@ -18,9 +20,13 @@ def time_since_last_unprocessed
...
   def week_hour_breakdown(column)
+    column.define_singleton_method(:quoted_id) do
+      column
+    end
+
     @scoped_model.
-      group("EXTRACT(dow FROM #{column})::integer").
-      group("EXTRACT(hour FROM #{column})::integer").
+      group("EXTRACT(dow FROM #{quote(column)})::integer").
+      group("EXTRACT(hour FROM #{quote(column)})::integer").
```
SteveMarshall: Does group not accept parameterised input? Annoying. This looks good, though.

lib/calculated_metrics.rb
```diff
@@ -44,32 +45,39 @@ def percent_rejected(prison, reason=nil)
     end
   end
 
-  def calculate_percentiles(column, percentile)
+  private
+  def calculate_percentiles(column, percentile) 
+    # This is a truly horrible hack. It lets us get past AR built-in value quoting, by pretending to be another object.
+    # It is needed so that we can parametrize two queries below by column, without resorting to putting the query together
```
SteveMarshall: Minor typo: `parametrize`.

lib/calculated_metrics.rb
```diff
@@ -44,32 +45,39 @@ def percent_rejected(prison, reason=nil)
...
+    # by hand. Don't call this method from outside this class, please.
```
SteveMarshall: I dunno if we need the last sentence: does the `private` not imply that?

SteveMarshall:
A few super-minor comments. Other than those, this looks great :+1:

lib/detailed_metrics.rb
jasiek: Nope.

PR #164, opened 2015-02-16
Experiment 1: Which copy discourages users from contacting us?

jasiek:


SteveMarshall:
Can switch between variants on the server side? People for whom the CSS doesn't load might end up seeing both, and that'd be not great.

jasiek:
We could, but this is less effort.

jasiek:
Also, the `.variant` element is styled within this page, so it doesn't matter if CSS loads or not.

SteveMarshall:
Sorry, I wasn't clear. People with CSS disabled will see both, people with JS disabled (or not loading for whatever reason) will probably see neither (and we can't track people with JS disabled).

If we're happy with that for this experiment, that's fine, but if we're going to be doing more experiments, we should think about how we can do it in a less exclusive way.

Also: are we passing through some sort of experiment ID to zendesk so we can see how this affects the emails we receive?

matharden:
Would be good to copy the original into a `<noscript>` for any user with no JavaScript

app/views/feedbacks/_form.html.haml
```diff
@@ -1,18 +1,38 @@
-%p If you need to ask a question about: 
+%script{src: "//www.google-analytics.com/cx/api.js?experiment=4uAdObjoRp2Cz3Ve100AvA"}
+:css
```
matharden: Remove this and use `.hidden`.

app/views/feedbacks/_form.html.haml
```diff
@@ -49,3 +69,11 @@
     .actions
       = f.submit 'Send', :class => 'button button-primary', :id => 'submit-button'
       
+:javascript
+  document.addEventListener('DOMContentLoaded', function() {
+    var variants = ['.variant.a', '.variant.b'];
+    var chosenVariant = cxApi.chooseVariation();
+    $(variants[chosenVariant]).show();
```
matharden: …and use `.removeClass('hidden')` here instead of `show` method.

PR #168, opened 2015-02-20
Better session timeout handling

jasiek:
If there's no session, redirect with a timeout message on both `edit` and `update` actions.

[#88771750]


matharden:
Looks good. But let's merge-Monday.

SteveMarshall:
:shipit: :boom:

PR #174, opened 2015-03-04
Each e-mail contains schema.org metadata about the booking event.

jasiek:
A button in Gmail that provides "add to calendar" functionality from email list view using schema.org.

### DIFF
--- a/app/views/mailers_common/_booking_confirmation.html.haml
+++ b/app/views/mailers_common/_booking_confirmation.html.haml
@@ -114,3 +114,6 @@
 
 %p
   %small Visit ID: #{@visit.visit_id}
+
+%script{type: 'application/ld+json'}
+  = event_schema_json(@visit, @slot)
--- a/app/helpers/visit_helper.rb
+++ b/app/helpers/visit_helper.rb
@@ -37,16 +37,16 @@ module VisitHelper
     mail_to prison_data['email']
   end
 
-  def prison_postcode
-    prison_data['address'][-1]
+  def prison_postcode(source=visit)
+    prison_data(source)['address'][-1]
   end
 
   def prison_slot_anomalies
     prison_data['slot_anomalies']
   end
 
-  def prison_address(glue='<br>')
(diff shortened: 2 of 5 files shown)


tonyduarte:
hey Jan, what does this address? thanks.

jasiek:
You know, how in GMail you will sometimes have a hint above an e-mail (https://support.google.com/calendar/answer/47802?hl=en) that prompts you to add it to the calendar? Me and @matharden thought it would be useful to have this for those of our clients who use GMail and other "smart" email providers.

SteveMarshall:
@jan Could you rebase to describe that reasoning in the commit message? (And, ultimately, the PR) That way, anyone who wants to see where this code came from gets the same description without too much effort.

app/views/mailers_common/_booking_confirmation.html.haml
```diff
@@ -114,3 +114,6 @@
 
 %p
   %small Visit ID: #{@visit.visit_id}
+
+%script{type: 'application/ld+json'}
```
matharden: I'm concerned about this. This puts a `script` tag in the email. I thought webmail clients strip these from email as a security measure. Does this work?

jasiek: That could be the case. Alternatively, we can use HTML tags to convey the same information.

SteveMarshall: We should do this in HTML if we can, I think. It makes our markup more semantic and means the metadata is then a first-class citizen and less likely to be subject to the sort of scrutiny that JavaScript is (by less modern mail clients, anyway), as @mat mentions.

matharden: Defo go for HTML!

SteveMarshall:
Any chance we can get this done and merged this week?

threedaymonk:
Closing stale PR as discussed. This feature would be nice to have, however.

PR #184, opened 2015-03-25
Added ping controller to display build / release information

lukaszraczylo:
Using format shared across other projects to display build information.

Objective: We should be able to monitor current version / release id on PVB project. 
@ashb pointed me to PQ project - https://github.com/ministryofjustice/parliamentary-questions/blob/dev/app/controllers/ping_controller.rb and proposed to follow that pattern.

### DIFF
--- a/app/controllers/ping_controller.rb
+++ b/app/controllers/ping_controller.rb
@@ -1,0 +1,8 @@
+class PingController  < ApplicationController
+  respond_to :json
+
+  def index
+    respond_with { :version_number => VERSION_INFO['version_number'], :build_date => VERSION_INFO['build_date'],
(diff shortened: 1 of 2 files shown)


SteveMarshall:
@lukaszraczylo Can you update the title/description (and the commit message) to give some context on why you need this as well as what it is? It's hard to know if this is the right solution (or appropriate for merging) without knowing what you're trying to solve by doing it.

SteveMarshall:
Looks good to me; thanks for clarifying the description, @lukaszraczylo! Does this depend on anything, or can it be merged now?

ashb:
This change is standalone - if it's missing the fields in the META it will just have null in the field.

SteveMarshall:
:bomb:…:boom:

PR #187, opened 2015-03-30
Experiment 1: Outcome B is preferrable.

jasiek:
[#88257056]

### DIFF
--- a/app/views/feedbacks/_form.html.haml
+++ b/app/views/feedbacks/_form.html.haml
@@ -1,49 +1,16 @@
 %script{src: "//www.google-analytics.com/cx/api.js?experiment=4uAdObjoRp2Cz3Ve100AvA"}
-%noscript
-  %p If you need to ask a question about: 
 
+%p
+  You can use this form to tell us about a problem you're having with the website. You must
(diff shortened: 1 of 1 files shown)


matharden:
Is this removing this the experiment?

jasiek:
Yes, it is.

app/views/feedbacks/_form.html.haml
```diff
@@ -1,37 +1,5 @@
...
-
-  %p
-    You must 
-    %a(href="http://www.justice.gov.uk/contacts/prison-finder") contact the prison
-    for help. 
-
-  %p You can use this form to tell us about a problem you’re having with the visit request website, or to ask us a question about how to use it. 
-
-.variant.b.hidden
+.experiment-1-outcome-b
```
matharden: As it's removing the experiment, could we remove this div element?

jasiek: I'll remove that. I thought it might be helpful to have that in the markup for later reference, but maybe that wasn't a good idea.

matharden:
Looks good.

PR #189, opened 2015-04-01
Visits are not processed on bank holidays

jasiek:
This adds an extra lead day for every bank holiday. Whether a bank holiday is bookable or not is still controlled by `unbookable_dates`.


spec/lib/schedule_spec.rb
```diff
@@ -51,19 +51,32 @@
...
+      # Su Mo Tu We Th Fr Sa
+      #           1  2 <3> 4
+      #  5 <6> 7  8  9 10 11
+      # 12 13 14 15 16 17 18
+      # 19 20 21 22 23 24 25
+      # 26 27 28 29 30
+
+      start_date = Date.new(2015, 4, 2)
+      subject.dates(start_date).first.should == start_date + 8
     end
```
SteveMarshall: It looks like this depends on `bank-holidays.json` containing bank holidays for this year? We should probably mock the bank holidays explicitly just in case they disappear for some reason.

jasiek: Not sure I understand what you mean. Can you give me an example?

SteveMarshall: Someone runs `rake update_bank_holidays` in the future, and that gives us a new `bank-holidays.json`. https://www.gov.uk/bank-holidays.json [isn't a supported API](https://insidegovuk.blog.gov.uk/2014/09/15/current-state-of-apis-on-gov-uk/) from GOV.UK, though, so they could well remove all past bank holidays from it with no notice.

Having these tests depend (albeit indirectly) on an unsupported API seems worrisome, particularly given we could just mock the data for this test (thus decoupling the tests from what is, really, live config). It feels, I suppose, like we're testing the data as much as the behaviour, rather than testing the latter in isolation.

We should probably have another test checking that the data is still in the format we want, too, but I can't find a JSON schema that's published anywhere for us to do that (which, I guess, ties back into the fact that it's an unsupported API)…

jasiek: How do you mean that it is an unsupported API?

SteveMarshall: I mean that GDS haven't committed to maintain that JSON file in the current form. They could remove it, change the format, or anything they like as they see fit with no notice. I'd like to decouple “the JSON format has changed” issues from “our code has broken” so we can identify that more easily.

jasiek: It seems that it is publicly accessible, so I'd imagine they'd think twice before disabling/changing it. I'd be in favour of a test that pulls it and checks the format - even a Jenkins task that does this once a week would be sufficient. Alternatively, we could switch to ICS - we could then pull the feed from a number of sources and not worry too much about the format.

SteveMarshall: I'm happy with the JSON—as you say, I think they'd think twice about changing it—but we should have a test to make sure we don't get caught out if they do.

The current test should still use mocks just in case the current data we rely on is removed from the JSON: I don't want to have to update the test if the data changes.

jasiek:
Is this good to go?

SteveMarshall:
Thanks for adding the tests for the format changes: they look great :+1: 

`it "assumes bookings are not processed on bank holidays"` still depends on Easter bank holidays for this year being in `config/bank-holidays.json`, though. We should mock those two dates so the test doesn't fail if they're removed from the JSON.

jasiek:
Ah, now it all make sense. Fix coming soon.

SteveMarshall:
Looks good to me! :boom:

PR #190, opened 2015-04-02
Collapse prison aliases into buckets based on NOMIS IDs

jasiek:
What problem does this solve? We have a number of prisons with different booking regimes that are actually the same prison (eg. Highpoint North/South) operationally. Given visits are processed by the same people, it makes sense to consolidate those for metric tracking purposes. I've adopted an approach where we'll place them into buckets based on their NOMIS IDs (since those are unique for each *actual* prison).

This PR contains a migration which may require ca 10 mins of downtime. Please don't merge yet.


SteveMarshall:
Can you add more detail on why we're making this change to the PR title/description/commit messages? Why do we want to collapse into NOMIS ID buckets? What benefit does that give us? That info might be useful for someone coming to this code cold when supporting it later…

app/helpers/application_helper.rb
```diff
@@ -51,4 +51,12 @@ def conditional_text(variable, prefix='', suffix='')
       prefix + variable.to_s + suffix
     end
   end
+
+  def prison_name_for_id(nomis_id)
+    @nomis_id_to_prison ||= \
+    Rails.configuration.prison_data.inject({}) do |h, (prison_name, prison_data)|
```
SteveMarshall: Can this line be joined up with the one above more clearly? I realise the `\` is intended to do that, but the combination of indentation and line-length make this harder to scan and see the relationship. Maybe just make it one line, like this:

```ruby
@nomis_id_to_prison ||= Rails.configuration.prison_data.inject({}) do |h, (prison_name, prison_data)|
```

Alternatively, put `Rails.configuration.prison_data` in a variable?

SteveMarshall:
Does removing the out-of-scope prisons from `prison_data_*.yml` have any impact on users? What happens if someone comes to our service and wants to book a visit for Hull, Acklington, or wherever else? Do they still get the “contact the prison” message as on production now, or are we removing that behaviour?

jasiek:
We're removing that behaviour.

jasiek:
Okay, updated - have a look now.

SteveMarshall:
Can you fix the test that's failing on CircleCI, or are you expecting it to fail for now? Could you also rebase and make the commits such that each represents a working, complete chunk of the work?

jasiek:
It's been rebased, I expect the CircleCI test to pass again.

spec/lib/schedule_spec.rb
```diff
@@ -187,16 +187,6 @@
...
-  context "prison without unbookable dates" do
-    let :prison do
-      Rails.configuration.prison_data['Oakwood'].dup
-    end
-
-    it "should return a list of dates" do
-      subject.dates(start_date = Date.new(2014, 12, 4)).first.should be_nil
-    end
-  end
-
```
SteveMarshall: Given that we're not removing unbookable prisons after all, can we add this back?

SteveMarshall:
Can you rebase this to group logical chunks of changes into commits, with messages describing why you've made each change? For example, all the renaming `prison` -> `nomis_id` in one commit (with tests passing at the end of each commit).

jasiek:
I've consolidated related commits as you requested.

SteveMarshall:
Thanks, that's clearer.

Can the removed test in `spec/lib/schedule_spec.rb` can be added back, or is it invalid now?

spec/lib/schedule_spec.rb
jasiek: Done

jasiek:
OK, once this is good to go I need to coordinate with Tony to roll this out as it incurs a short downtime.

SteveMarshall:
Looks good to go now. Thanks! :boom:

PR #192, opened 2015-04-13
Lindholme - May bank holiday time changes

RoseFaulkner:
May holiday time changes requested by Lindholme.

015-05-04:
    - 1345-1600
2015-05-25:
    - 1345-1600


SteveMarshall:
@RoseFaulkner Can you add the reason for this change to your description, and the name of the prison it affects to the title? (For example, was this requested by Lindholme, or have we done this pre-emptively?)

PR #194, opened 2015-04-22
HOT FIX: Instantiate the new Schedule object correctly so we can book again

ashb:
This is alnost certainly not the right way of doing the fix, but it
works and means we can book visits again.

Whithout this fix we were getting no slots returned.


ashb:
Tested by deploying this to staging and slots show up agian.

PR #195, opened 2015-04-23
Per Webops' requests, this exposes a /healthcheck.json endpoint

jasiek:


lib/sendgrid_helper.rb
```diff
@@ -24,4 +24,23 @@ def self.handle_response(url)
...
+    false
+  end
+
+  def self.messagelabs_alive?(host='cluster2.gsi.messagelabs.com')
+    smtp_alive?(host, 25)
+  end
+
+  def self.sendgrid_alive?(host='smtp.sengrid.net')
+    smtp_alive?(host, 587)
+  end
```
ashb: It looks like these two functions `messagelabs_alive?` and `sendgrid_alive?` aren't used so we should remove them

spec/lib/sendgrid_helper_spec.rb
```diff
@@ -61,4 +61,41 @@
...
+      'smtp.sendgrid.net'
+    end
+
+    let :port do
+      587
+    end
+
+    context "when it connects" do
+      it "will return true" do 
+        SendgridHelper.smtp_alive?(host, port).should be_true
```
ashb: Does this test hit the actual SMTP server? That's probably bad form to have our tests open a socket to someone elses mail server every time we run tests.

jasiek:
Fixed, @ashb

SteveMarshall:
This looks good to me, save @ashb’s [comments on `lib/sendgrid_helper.rb`](#discussion_r29046291). Do we need those two methods?

lib/sendgrid_helper.rb
jasiek: Ah, right, I'll remove them.

SteveMarshall:
@jasiek Just had another look at this; can we (in a new PR, of course) make it 500 when any of the services return false? Sorry! Should have spotted that when reviewing.

PR #199, opened 2015-04-28
Prevent a cancellation form from showing when not following a link from ...

jasiek:
...a confirmation e-mail.

[#92999494]


SteveMarshall:
Can you update the commit and the PR to explain why we need this change? What problem is this fixing?

matharden:
I'm not sure this fixes the problem. When can the user see the cancellation form (ie. when is the `state` provided)?

matharden:
As just discussed. I'm rejecting this.

PR #200, opened 2015-04-29
When a visitor wants to cancel a visit and forgets to check a checkbox, the app doesn't blow up.

jasiek:


SteveMarshall:
Do we need to change the outcomes from symbols to strings throughout? The only reason I can see (but I may be missing something…) is to remove a `to_sym` in `MetricsLogger#visit_status` and to use `%w{}` instead of `[:whatever, :other_thing]` in `VisitMetricsEntry`?

jasiek:
Yup. Strange, I thought I made that change...

jasiek:
When you save a symbol to the DB and then retrieve it, you get a string. It's easier to just use strings every time.

spec/lib/metrics_logger_spec.rb
```diff
@@ -109,15 +111,25 @@
...
+      subject.visit_status(sample_visit.visit_id).should == reason
+    end
+  end
+
+  it "blows up when trying to cancel with an unknown reason" do
+    expect {
+      subject.record_visit_request(sample_visit)
+      subject.record_link_click(sample_visit)
+      subject.record_booking_cancellation(sample_visit.visit_id, 'whatever')
+    }.to raise_error(ActiveRecord::RecordInvalid)
```
SteveMarshall: Are we handling this cleanly elsewhere, or is this a valid situation for a 500 error?

SteveMarshall:
Ah, yeah, of course it does. Fair enough.

spec/lib/metrics_logger_spec.rb
jasiek: No, because it's not supposed to happen. It's a sanity check for when someone adds a new outcome but forgets to add it to a list of valid outcomes in the model definition.

SteveMarshall: :+1:

app/views/visit/_cancellation_form.html.haml
```diff
@@ -3,9 +3,9 @@
...
 
     %label.block-label(for="cancel_booking")
-      %input#cancel_booking{ type: "checkbox", value: "#{type}", name: "cancel" }
+      %input#cancel_booking{ type: "checkbox", value: true, name: "cancel" }
       = hidden_field_tag :state, @state
       Yes, I want to cancel this visit
 
   .actions
-    %input.button.button-primary{ type: "submit", value: "Cancel #{type}" }
+    %input.button.button-primary{ type: "submit", value: "Cancel" }
```
matharden: This means the user doesn't see they are cancelling the "request" or "visit" where appropriate. I liked how we previously made that distinction.

SteveMarshall: :+1:

spec/lib/metrics_logger_spec.rb
matharden: :+1:

app/views/visit/_cancellation_form.html.haml
jasiek: I agree that it was nicer. But is it really that important?

app/controllers/visit_controller.rb
```diff
@@ -19,13 +19,17 @@ def update_status
     @visit = encryptor.decrypt_and_verify(params[:state])
 
     if params[:cancel]
-      PrisonMailer.booking_cancellation_receipt_email(@visit).deliver
-      metrics_logger.record_booking_cancellation(params[:id], "#{params[:cancel]}_cancelled")
+      if @visit_status.nil?
```
matharden: :+1:

matharden:
Please could you add a test to make sure an email is only sent when a 'visit' is cancelled and not a 'request'?

app/views/visit/_cancellation_form.html.haml
matharden: The heading above the form does mention 'request' when appropriate. This PR is not a design change, so I'd prefer to retain this as previously designed.

jasiek: OK, I'll change that back.

jasiek:
Is this good to go?

SteveMarshall:
I'm happy with it. @matharden? And have @tonyduarte or @RoseFaulkner tested it?

matharden:
Looks good to me. Let's put it on pre-prod first thing tomorrow and I'll run through it.

PR #207, opened 2015-05-14
Ensure healthcheck 5XXs on poor health

SteveMarshall:
When services we depend on are down (Sendgrid, Messagelabs, and the database), the healthcheck endpoint should return a HTTP 5XX response so that Pingdom fires alerts when things go awry.

### DIFF
--- a/app/controllers/heartbeat_controller.rb
+++ b/app/controllers/heartbeat_controller.rb
@@ -2,12 +2,14 @@ class HeartbeatController < ApplicationController
   permit_only_with_key
 
   def healthcheck
-    render json: {
-      checks: {
-        sendgrid: sendgrid_alive?,
-        messagelabs: messagelabs_alive?,
-        database: ActiveRecord::Base.connection.active?
-      }
+    checks = {
+      sendgrid: sendgrid_alive?,
+      messagelabs: messagelabs_alive?,
+      database: ActiveRecord::Base.connection.active?,
+    }
+    status = :bad_gateway unless checks.values.all?
+    render status: status, json: {
+      checks: checks
     }
   end
 
--- a/app/controllers/heartbeat_controller.rb
+++ b/app/controllers/heartbeat_controller.rb
@@ -2,10 +2,16 @@ class HeartbeatController < ApplicationController
   permit_only_with_key
 
   def healthcheck
+    database_active = begin
+      ActiveRecord::Base.connection.active?
+    rescue PG::ConnectionBad
+      false
+    end
+
     checks = {
       sendgrid: sendgrid_alive?,
       messagelabs: messagelabs_alive?,
-      database: ActiveRecord::Base.connection.active?,
+      database: database_active,
     }
     status = :bad_gateway unless checks.values.all?
     render status: status, json: {
--- a/spec/controllers/heartbeat_controller_spec.rb
+++ b/spec/controllers/heartbeat_controller_spec.rb
@@ -30,6 +30,43 @@ describe HeartbeatController do
         parsed_body['checks']['sendgrid'].should be_true
         parsed_body['checks']['database'].should be_true
       end
+
+      it "errors if messagelabs is down" do
+        PrisonMailer.should_receive(:smtp_settings).and_return(address: host = double, port: port = double)
+        SendgridHelper.should_receive(:smtp_alive?).with(host, port).once.and_return(false)
+        VisitorMailer.should_receive(:smtp_settings).and_return(address: host = double, port: port = double)
+        SendgridHelper.should_receive(:smtp_alive?).with(host, port).once.and_return(true)
+        get :healthcheck
+
+        assert_response(:bad_gateway, response.status)
+        parsed_body = JSON.parse(response.body)
+        parsed_body['checks']['messagelabs'].should be_false
+      end
+
+      it "errors if sendgrid is down" do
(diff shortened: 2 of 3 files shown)


app/controllers/heartbeat_controller.rb
```diff
@@ -2,11 +2,19 @@ class HeartbeatController < ApplicationController
   permit_only_with_key
 
   def healthcheck
-    render json: {
+    messagelabs_alive = messagelabs_alive?
+    sendgrid_alive = sendgrid_alive?
+    database_alive = ActiveRecord::Base.connection.active?
+    status = status = :bad_gateway unless [
```
jasiek: Why the double assignment?

spec/controllers/heartbeat_controller_spec.rb
```diff
@@ -2,27 +2,70 @@
...
       it "talks to sendgrid and messagelabs" do
         PrisonMailer.should_receive(:smtp_settings).and_return(address: host = double, port: port = double)
-        SendgridHelper.should_receive(:smtp_alive?).with(host, port).once
+        SendgridHelper.should_receive(:smtp_alive?).with(host, port).once.and_return(true)
         VisitorMailer.should_receive(:smtp_settings).and_return(address: host = double, port: port = double)
-        SendgridHelper.should_receive(:smtp_alive?).with(host, port).once
+        SendgridHelper.should_receive(:smtp_alive?).with(host, port).once.and_return(true)
         get :healthcheck
+
+        assert_response(:success, response.status)
```
jasiek: Does the following work?

`response.status.should be :success`

app/controllers/heartbeat_controller.rb
SteveMarshall: Arg. Typo. I've also just realised there's a nicer way to do this whole block using a hash with `.values.any?`.

spec/controllers/heartbeat_controller_spec.rb
```diff
@@ -2,27 +2,70 @@
...
+      end
+
+      it "errors if messagelabs is down" do
+        PrisonMailer.should_receive(:smtp_settings).and_return(address: host = double, port: port = double)
+        SendgridHelper.should_receive(:smtp_alive?).with(host, port).once.and_return(false)
+        VisitorMailer.should_receive(:smtp_settings).and_return(address: host = double, port: port = double)
+        SendgridHelper.should_receive(:smtp_alive?).with(host, port).once.and_return(true)
+        get :healthcheck
+
+        assert_response(:bad_gateway, response.status)
```
jasiek: I think these three lines could be wrapped into a helper.

spec/controllers/heartbeat_controller_spec.rb
SteveMarshall: Unfortunately not.

Because we're on rspec 2, expectations don't map between response symbol names and numeric status codes, and I thought it nicer to use the symbol, even though it does mean we're not being quite so rspec-native. Waddya think?

jasiek: It's fine.

SteveMarshall:
@jasiek I've refactored a bit, and resolved the two issues you mentioned. Let me know what you think and I'll look at any more changes you think this needs when I'm back on Monday.

app/controllers/heartbeat_controller.rb
```diff
@@ -2,12 +2,14 @@ class HeartbeatController < ApplicationController
...
-        messagelabs: messagelabs_alive?,
-        database: ActiveRecord::Base.connection.active?
-      }
+    checks = {
+      sendgrid: sendgrid_alive?,
+      messagelabs: messagelabs_alive?,
+      database: ActiveRecord::Base.connection.active?,
+    }
+    status = :bad_gateway unless checks.values.all?
+    render status: status, json: {
```
jasiek: Does `render status: nil` do anything?

jasiek:
LG

app/controllers/heartbeat_controller.rb
SteveMarshall: Yeah: it's treated the same as not passing `status`, so results in a `200 Success` response, which is exactly what we want.

SteveMarshall:
It would appear, on testing on preprod, this doesn't catch the database errors properly so I need to do a bit more work to do that.

app/controllers/heartbeat_controller.rb
```diff
@@ -2,12 +2,20 @@ class HeartbeatController < ApplicationController
...
   def healthcheck
-    render json: {
-      checks: {
-        sendgrid: sendgrid_alive?,
-        messagelabs: messagelabs_alive?,
-        database: ActiveRecord::Base.connection.active?
-      }
+    database_active = begin
+      ActiveRecord::Base.connection.active?
+    rescue PG::ConnectionBad
```
SteveMarshall: I'm not entirely happy with this dependency on `PG` directly: it could make switching DBs down the line a bit harder. @jasiek / @threedaymonk any thoughts on another class of error I could rescue here?

Also, is there a cleaner way to do this whole block?

jasiek: It's fine. If anything we'll get rid of the database completely when we integrate with the API, so no need to worry about it IMO.

SteveMarshall:
Right, I've tested this works in preprod: Pingdom picks up the database being offline, fires an alert to PagerDuty, and that fires alerts accordingly.

Issue #209, opened 2015-05-18
Visit helper


mattmb:
Somehow prison_data is sometimes nil in this view?
http://sentry.prod-prisonvisits.dsd.io/default/default/group/2380
```
ActionView::Template::Error
undefined method `[]' for nil:NilClass
/app/helpers/visit_helper.rb:53:in `adult_age'
/app/views/shared/visitors_details/_visitor_fields.html.haml:4:in 
```

Issue #208, opened 2015-05-18
Unbookable dates


mattmb:
There seems to be a bug regarding unbookable dates:
http://sentry.prod-prisonvisits.dsd.io/default/default/group/2308/

```
NoMethodError
undefined method `[]' for nil:NilClass
/lib/schedule.rb:7:in `initialize'
/app/controllers/deferred/slots_controller.rb:8:in `new'
/app/controllers/deferred/slots_controller.rb:8:in `edit'
```

mattmb:
So after investigation:
https://github.com/ministryofjustice/prison-visits/blob/master/app/controllers/deferred/slots_controller.rb#L9
references `visit.prisoner.prison_name` and if this is nil we get this exception.

One way the user can cause this to happen:
* Start a new session, fill in the service.
* On the review page click edit against the prisoner details.
* Blank the prison name, press continue.
* You will get an error message. However the prison name is cleared from the session.
* Use back/forward buttons to navigate to the slot picker.

matharden:
Do you agree this issue is now resolved? I've raised a new issue for the user side of this.

mattmb:
Yep happy to track it in a new issue now that we know the root cause :smile:

PR #210, opened 2015-05-20
Raise useful exception when prison not found

threedaymonk:
In issues #208 and #209, we see exceptions raised when there is no prison data for the prison name supplied. It's more helpful to raise an exception earlier that tells us that the key does not exist in the prison data hash.

Instead of:

    ActionView::Template::Error: undefined method `[]' for nil:NilClass

the exception raised by an unknown prison name is now:

    KeyError: key not found: "bar"

This should assist us in tracking down the root cause of these errors.


mattmb:
LGTM :+1:

PR #211, opened 2015-05-21
Cardiff - blocked date - June 3rd 2015

RoseFaulkner:
Cardiff requested that Wed, June 3rd is blocked off as there will be no
visits on that day.


SteveMarshall:
This looks good, @RoseFaulkner. Has it been verified on staging?

PR #212, opened 2015-05-21
Improve environment

threedaymonk:
This contains a few things to improve my experience of working on this project.

* Changes the local database configuration slightly in order to permit the use of PostgreSQL peer authentication (so that the specs run on Linux without additional configuration). I don't believe that this will cause any problems, as this is the configuration used in other projects.
* Silences the useless list of all spec files vomited out when running `rake`. This makes it easier to see what's actually happening.
* Adds Simplecov to measure how much of the codebase is not exercised when running tests, so that we can identify and mitigate risks when making changes.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,4 +32,6 @@ group :test, :development do
   gem 'chromedriver-helper'
   gem 'database_cleaner'
   gem 'rspec-html-matchers'
+  gem 'simplecov', '~> 0.7.1', require: false
+  gem 'simplecov-rcov', require: false
 end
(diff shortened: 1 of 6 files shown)


Gemfile
```diff
@@ -18,6 +18,8 @@ gem 'prison_staff_info', git: 'git@github.com:ministryofjustice/prison_staff_inf
 gem 'pg'
 gem 'statsd-ruby', require: 'statsd'
 gem 'curb'
+gem 'simplecov', '~> 0.7.1', require: false
+gem 'simplecov-rcov', require: false
```
SteveMarshall: Should these be in the `:test, :development` group below?

threedaymonk: Argh, yes, you're right.

threedaymonk: Fixed.

PR #213, opened 2015-05-21
Clean code and improve coverage

threedaymonk:
* Removes some code that is never used (as established by Simplecov – see #212 – and inspection).
* Improves coverage from about 91 to 96%.
* Makes some small improvements to structure and readability where coverage was adequate to give confidence.

The most notable removal is of the legacy visit migrations. This was implemented for a structural change in December last year. Five months on, we can be confident that we won't need to process any more of these legacy requests, and this code is now superfluous.

### DIFF
--- a/app/models/confirmation.rb
+++ b/app/models/confirmation.rb
@@ -16,6 +16,16 @@ class Confirmation
   validate :unlisted
   validate :banned
 
+  def slot_selected?
+    [0, 1, 2].include?(slot)
+  end
+
+  def slot
+    outcome && outcome.starts_with?('slot') && outcome.gsub('slot_', '').to_i
+  end
+
+private
+
   def check_outcome
     outcomes = [
       NO_VOS_LEFT,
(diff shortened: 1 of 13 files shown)


app/models/confirmation.rb
```diff
@@ -1,6 +1,8 @@
 class Confirmation
   include ActiveModel::Model
-  attr_accessor :outcome, :message, :vo_number, :no_vo, :no_pvo, :renew_vo, :renew_pvo, :banned_visitors, :unlisted_visitors, :visitor_not_listed, :visitor_banned, :canned_response, :closed_visit
+  attr_accessor :outcome, :message, :vo_number, :no_vo, :no_pvo, :renew_vo,
+    :renew_pvo, :banned_visitors, :unlisted_visitors, :visitor_not_listed,
+    :visitor_banned, :canned_response, :closed_visit
```
SteveMarshall: I wonder if it might be nicer, for future diffing purposes, to just break this line after every `,`. That said, that'll create an incredibly long line. Happy either way, but possibly worth considering.

threedaymonk: As you say, I think that would lead to a very long list at the top of the file, which might just be distracting. I don't think it's likely to be particularly volatile.

SteveMarshall: :+1:

dotemacs:
Looks fine, I'll merge.

PR #214, opened 2015-05-26
Decouple prison-specific ID requirements from Rails internals

threedaymonk:
The use of `lookup_context` is highly dependent on the internal workings of Rails, and makes it hard to upgrade to Rails 4.2.

We replace this with a more explicit method:

* Find the template (either specific or generic) on disk
* Return its content
* Render as Markdown if appropriate, using separate text and HTML templates

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -62,4 +62,17 @@ module ApplicationHelper
     end
     @nomis_id_to_prison[nomis_id]
   end
+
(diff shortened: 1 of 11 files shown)


dotemacs:
Looks good. Merging.

SteveMarshall:
This is *great*, nice work @threedaymonk! The simplicity of `prison_specific_id_requirements` in  :boom:

PR #221, opened 2015-06-19
Ensure datepicker specs run after 2015/06/09

SteveMarshall:
`datepicker_spec` `Choosing a date in the past` failed for two reasons:

- The `Timecop.travel` call was *after* the page was rendered, so had no impact on the rendering.
- Travelling to the past resulted in cookies that expired in the past being issued to the test browser. Because the test browser doesn't respect our time-travelling, it considered these cookies expired instantly.

To avoid the session expiring, we should always travel to a suitable date in the future (a Tuesday, so we can test booking a prior working day of the week).

Also ensure we pass the server's idea of “today” to the JavaScript to avoid weird miscalculations of end-dates on the client-side.

### DIFF
--- a/spec/features/datepicker_spec.rb
+++ b/spec/features/datepicker_spec.rb
@@ -3,6 +3,14 @@ require 'browserstack_helper'
 RSpec.feature "visitor selects a date" do
   include_examples "feature helper"
 
+  before :all do
+    Timecop.travel(Time.now.next_week(:tuesday).at_noon)
+  end
+
+  after :all do
+    Timecop.return
+  end
+
   before :each do
     allow_any_instance_of(VisitController).to receive(:metrics_logger).and_return(MockMetricsLogger.new)
     allow_any_instance_of(EmailValidator).to receive(:validate_dns_records)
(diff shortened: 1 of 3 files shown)


spec/features/datepicker_spec.rb
```diff
@@ -3,6 +3,15 @@
 RSpec.feature "visitor selects a date" do
   include_examples "feature helper"
 
+  before :all do
+    next_tuesday = (Time.new + 1.week).beginning_of_week.at_noon + 1.day
```
threedaymonk: Does `beginning_of_week` always give us a Monday, or does it change with locale?

threedaymonk: It *is* variable: http://api.rubyonrails.org/classes/DateAndTime/Calculations.html#method-i-beginning_of_week

I'd suggest using `beginning_of_week(:monday)`, though it wouldn't surprise me if this assumption is pervasive.

spec/features/datepicker_spec.rb
```diff
@@ -76,11 +82,9 @@
 
     scenario 'Choosing a date in the past' do
       # A Tuesday, so that yesterday is a working day
```
threedaymonk: This comment refers to a deleted line.

PR #223, opened 2015-06-22
Clarify intent of schedule tests

SteveMarshall:
We shouldn't have to do date maths in our head to understand the tests, so remove the magic numbers. 

It also turns out that the code doesn't behave at all as it should, so this change is purely to clarify what the code *actually* does. That way, when we come to fix the behaviour in future there's a good example of how to communicate and test that change.

Also clarify some parameters and make date construction consistent.

### DIFF
--- a/spec/lib/schedule_spec.rb
+++ b/spec/lib/schedule_spec.rb
@@ -2,7 +2,8 @@ require 'rails_helper'
 
 RSpec.describe Schedule do
   subject do
-    Schedule.new(prison, [Date.new(2014, 8, 25), Date.new(2015, 4, 3), Date.new(2015, 4, 6)])
+    bank_holidays = [Date.new(2014, 8, 25), Date.new(2015, 4, 3), Date.new(2015, 4, 6)]
+    Schedule.new(prison, bank_holidays)
   end
 
   let :start_date do
@@ -13,7 +14,7 @@ RSpec.describe Schedule do
     let :prison do
       Rails.configuration.prison_data['Durham']
     end
-    
+
     it "returns days within 28 days, excluding the lead days" do
       subject.dates(start_date).each do |date|
         expect(date).to be >= start_date + 3
@@ -21,10 +22,10 @@ RSpec.describe Schedule do
       end
     end
 
-    it "rejects unbookable dates" do
+    it "doesn't offer unbookable dates" do
       subject.dates(start_date).each do |date|
-        expect(date).not_to eq(Date.parse('2014-12-25'))
-        expect(date).not_to eq(Date.parse('2014-12-26'))
+        expect(date).not_to eq(Date.new(2014, 12, 25))
+        expect(date).not_to eq(Date.new(2014, 12, 26))
       end
     end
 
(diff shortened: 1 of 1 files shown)


spec/lib/schedule_spec.rb
```diff
@@ -21,10 +22,10 @@
       end
     end
 
-    it "rejects unbookable dates" do
+    it "doesn't offer unbookable dates" do
       subject.dates(start_date).each do |date|
-        expect(date).not_to eq(Date.parse('2014-12-25'))
-        expect(date).not_to eq(Date.parse('2014-12-26'))
+        expect(date).not_to eq(Date.new(2014, 12, 25))
```
matharden: I actually find the .parse syntax easier to read.

SteveMarshall: I don't mind either way, I just changed this to match the rest of the file because this was the only use of parse.

spec/lib/schedule_spec.rb
```diff
@@ -44,26 +45,21 @@
...
-      start_date = Date.new(2014, 8, 23) # Saturday
-      expect(subject.dates(start_date).first).to eq(start_date + 6) # Friday
-
-      start_date = Date.new(2014, 8, 24) # Sunday
-      expect(subject.dates(start_date).first).to eq(start_date + 5) # Friday
+      week_start = monday = Date.new(2014, 8, 18)
+      first_bookable_visits = {
+        :monday => week_start.next_day(Date::DAYS_INTO_WEEK[:friday]),
+        :tuesday => week_start.next_day(Date::DAYS_INTO_WEEK[:saturday]),
+        :wednesday => week_start.next_week(:wednesday),
```
matharden: Durham have visits every day. When it's Wednesday there are 4 whole working days before next Wednesday. Should the next possible bookable day be Tuesday?

spec/lib/schedule_spec.rb
```diff
@@ -44,26 +45,21 @@
...
+        :thursday => week_start.next_week(:thursday),
+        :friday => week_start.next_week(:friday),
+        :saturday => week_start.next_week(:friday),
+        :sunday => week_start.next_week(:friday),
+      }
+
+      first_bookable_visits.each do |request_day, visit_on|
```
matharden: :+1:

spec/lib/schedule_spec.rb
```diff
@@ -87,7 +83,7 @@
 
     it "rejects days without slots" do
       subject.dates(start_date).each do |date|
-        expect(date.wday).not_to eq(5)
+        expect(date.friday?).to be false
```
matharden: :+1:

spec/lib/schedule_spec.rb
SteveMarshall: Probably, but [these sorts of issues]( are exactly what led me to look at this code in the first place. It turns out that the code doesn't behave at all as we intend, so this change is purely to clarify what the code *actually does*, so when we come to make the change in future, there's a good example of how to communicate that change in intent.

I should put that in the commit message and PR description! (I will, when you're done.)

spec/lib/schedule_spec.rb
matharden: OK. Fair enough.

spec/lib/schedule_spec.rb
matharden: Coolbeanz.

matharden:
Happy to merge this.

SteveMarshall:
@matharden Pushed with amended commit message, as discussed. Merge when ready!

matharden:
coolbeanz

PR #226, opened 2015-07-03
Update redcarpet to avoid CVE-2015-5147

SteveMarshall:
[CVE-2015-5147](https://security-tracker.debian.org/tracker/CVE-2015-5147) allows a stack overflow in header_anchor. That's bad. Don't allow that.


dotemacs:
Intense.

PR #230, opened 2015-07-16
Fix prison schedules to coincide with updated prison confirmation dates

dan-palmer:
#### Fix prison schedules to coincide with updated prison confirmation dates

**Step one: Encapsulate prison data in a model** :tada: 
This provides a class that wraps a prisons data as set in the prison config file, exposing methods to query it.
It exposes a class method for performing lookups based on a Prisons name.
The interface for an instance of the said model currently only exposes methods that are required for schedule related queries(e.g checking if a given day is one in which a prison is expected to be working).

**Step two: Encapsulate prison day** :tada: 
Decorate a given `Date` instance with prison specific predicates to aid in determining if the day is:
- a working day - help determine the confirmation email day
- a day available for visiting

**Step three: Enable scheduler to make use of the new functionality** :tada:
Provide two methods for retrieving the range of available dates for visitation & a single date for the expected confirmation email date.

 **Step four: Update old interface method calls** :tada: 
Update old scheduler interactions to utilise new interface.


spec/models/prison_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+  describe '.unbookable_dates' do
+    context 'when a prison has unbookable dates' do
+      let(:expected_dates) { Set.new [Date.new(2015, 7, 29), Date.new(2015, 12, 25)] }
+
+      it 'returns a set containing the dates' do
+        expect(subject.unbookable_dates).to eq expected_dates
+      end
+    end
+    context 'when a prison does not have unbookable dates' do
+      subject { constructor_for mock_prison_data.except 'unbookable' }
```
SteveMarshall: This is really neat :+1:

spec/models/prison_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+    end
+  end
+
+  describe '.works_weekends?' do
+    context 'when a prison has a works weekends boolean flag set' do
+      it 'returns that boolean' do
+        expect(subject.works_weekends?).to be_truthy
+      end
+    end
+    context 'when no booking window has been set' do
```
SteveMarshall: I'm not sure you meant for this to refer to booking window? :wink:

spec/models/prison_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+      subject { constructor_for mock_prison_data.except 'works_weekends' }
+
+      specify{ expect(subject.works_weekends?).to be_falsey }
```
SteveMarshall: Any reason you used `specify` rather than `it`?

spec/models/prison_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+    it 'returns the prisons slots as a hash' do
+      expect(subject.visiting_slots).to eq mock_slots_data
+    end
+  end
+
+  describe '.visiting_slot_days' do
+    let(:expected_slot_days) { %w<mon tue wed thu fri sat sun> }
+    it 'returns the abbreviated day names of available visiting days' do
+      expect(subject.visiting_slot_days).to eq expected_slot_days
+    end
```
SteveMarshall: I wonder if it's worth testing that this works with another set of visiting slot days?

spec/models/prison_spec.rb
dan-palmer: Well spotted! Bit of a copy and paste job for that spec me thinks. :+1:

spec/models/prison_spec.rb
dan-palmer: This is purely for the output it produces when the tests are run.

screen shot 2015-07-17 at 10 22 25

I think within the context of the spec itself & reading the code its still sensible. What do you think?

SteveMarshall: Nah, that's great: I was just curious what inspired the choice!

spec/models/prison_spec.rb
dan-palmer: cheers :smile:

spec/lib/prison_day_spec.rb
```diff
@@ -0,0 +1,226 @@
...
+        end
+      end
+    end
+
+    context 'on a public holiday' do
+      let(:bank_holidays) { [Date.new(2015, 8, 31)] }
+      before { stub_const("#{described_class.name}::BANK_HOLIDAYS", bank_holidays) }
+      subject { described_class.new(bank_holidays.first, prison_from(prison_data)) }
+
+      specify { expect(subject.staff_working_day?).to be false }
```
SteveMarshall: These (and `visiting_day?` later) might be slightly easier to read if you made them into shared examples? Something like this:

```ruby
it_behaves_like "a working day"
# or
it_behaves_like "a non-working day"
```

Dunno; waddya think?

spec/lib/prison_day_spec.rb
```diff
@@ -0,0 +1,226 @@
...
+    end
+
+    let(:bank_holiday_friday) { DAYS.fetch :friday }
+    let(:stub_bank_holidays) do
+      stub_const("#{described_class.name}::BANK_HOLIDAYS", [bank_holiday_friday])
+    end
+
+    let(:unbookable_monday) { DAYS.fetch :monday }
+
+    context 'a day registered as available for visitation for a given prison' do
```
SteveMarshall: The contexts for `.staff_working_day?` above all start with `on`, which makes the tests much clearer in the output; add those throughout for `.visiting_day?`, too, maybe?

spec/lib/prison_day_spec.rb
```diff
@@ -0,0 +1,226 @@
...
+      specify do
+        DAYS.except(:thursday).each do |day_name, date|
+          described_class.new(date, prison_with_visits_except_thursday).tap do |prison_day|
+            expect(prison_day.visiting_day?).to be true
+          end
```
SteveMarshall: I don't see a test for the inverse of this?

SteveMarshall: Oh, no, [found it](#diff-5756e50c8d509bf25b30f1754820e159R165).

spec/lib/prison_day_spec.rb
dan-palmer: Going to suggest we keep the contexts together for the time being so the spec doesn't end up scattered and hard to reason about. However, as we progress perhaps we'll introduce factories for generating different prisons with specific characteristics and we can tidy this up a bit. Something to put on the radar anyways.

SteveMarshall: :+1:

spec/lib/prison_schedule_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+    describe 'the range of dates covered' do
+      let(:the_day_after_the_conformation_email) { subject.confirmation_email_date.tomorrow }
+      let(:booking_window_days) { 28.days }
+      let(:the_last_bookable_day) { the_day_after_the_conformation_email + booking_window_days }
+
+      it 'starts from the day after the confirmation email' do
+        expect(subject.available_visitation_dates.first).to eq the_day_after_the_conformation_email
+      end
+
+      it 'the end date is equivalent to the number of booking window days for the prison plus the confirmation email date' do
```
SteveMarshall: How about calling this `it "doesn't offer booking windows past the number of bookable days after the confirmation email date"`?

SteveMarshall:
This is looking great :fireworks:

spec/lib/prison_schedule_spec.rb
dan-palmer: Great suggestion, this line doesn't read well at all :+1:

spec/helpers/visit_helper_spec.rb
```diff
@@ -84,7 +84,9 @@ def visit
     end
 
     it "provides a formatted date for when a response may be sent out" do
-      expect(helper.when_to_expect_reply(Date.parse("2014-10-03"))).to eq("Monday  6 October")
+      Timecop.travel(Date.parse("2014-10-03")) do
+        expect(helper.when_to_expect_reply).to eq("Friday  3 October")
```
SteveMarshall: Might be worth seeing if we can kill this probably-pointless space? @matharden: any idea why there'd be a double-space in the output here?

matharden: This is because the Rails date helper places a white space where the tens-digit would be for single digit numbers. Which is rubbish. I did put `gsub` fixes into views to get round this. Which I'm assuming has been missed (or removed) for this particular view, judging by this spec.

SteveMarshall: That's because this is directly testing the output of the helper and not touching the view. @dan-palmer anything we can do to get around the issue @matharden's describing and get rid of the stupid extra space? (If it's a hassle, don't bother!)

matharden: Something like `d.strftime("%A") + ' ' + d.strftime("%e %B").strip` on [ApplicationHelper](https://github.com/ministryofjustice/prison-visits/blob/fix/prison-schedules-and-prison-working-days/app/helpers/application_helper.rb#L14), but more elegant.

dan-palmer: Good shout @matharden. Exactly what we want. :+1:

SteveMarshall: Did you wanna fix that here, @dan-palmer, or in another PR?

dan-palmer: @SteveMarshall Yeah, I'll handle this in another PR

SteveMarshall: :+1:

SteveMarshall:
We should get @tonyduarte / @ThugRose to check this on preprod after #233 is done.

SteveMarshall:
:boom:

PR #233, opened 2015-07-24
Fixes missing `prison_name` on prisoner-details page

dan-palmer:
The `prison_name` helper method had accidentally been removed from the codebase in #213 and was causing errors when trying to render templates that require it - specifically when displaying reasons why a user cannot make a booking with a prison.

- Adds it back in with a test
- Adds a number of tests to the `/prisoner` page feature asserting the different error messages a user receives for different prison states (disabled, coming soon, IT issues)
- Refactors the feature test slightly to make it a bit more readable

screen shot 2015-07-24 at 16 53 04

### DIFF
--- a/spec/features/prison_information_page_spec.rb
+++ b/spec/features/prison_information_page_spec.rb
@@ -11,7 +11,8 @@ RSpec.feature 'visitor entering prisoner information' do
         context 'when a prison is not selected' do
           it 'displays an error message' do
             click_button 'Continue'
-            expect(page).to have_css("label[for='prisoner_prison_name'] .validation-message")
+
+            expect(page).to have_css "label[for='prisoner_prison_name'] .validation-message"
           end
         end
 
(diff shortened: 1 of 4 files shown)


spec/features/prison_information_page_spec.rb
```diff
@@ -1,43 +1,74 @@
...
+          let(:a_prison_with_it_issues) { 'Blantyre House' }
 
-          enter_prisoner_information(flow)
+          it 'displays a message about the prison not being available just yet' do
+            set_prison_to a_prison_with_it_issues
+            click_button 'Continue'
 
-          expect(page).to have_content('Visitor 1')
+            expect(page).
+              to have_content 'HMP Blantyre House is unable to process online visit requests right now.'          end
```
SteveMarshall: The formatting on this line is a bit odd. Is that intentional?

dan-palmer: The `end` was off screen in my editor :cry: - FP'd it onto the correct line. Well spotted.

PR #236, opened 2015-07-31
Stop old IEs choking on PDFs

SteveMarshall:
Upgrade prison_staff_info to a version that removes `Cache-Control: no-cache` from static assets to avoid [an issue over HTTPS, as detailed by Microsoft](https://support.microsoft.com/en-gb/kb/323308).

[Delivers #98948456]


SteveMarshall:
I've validated this on IE8 on Quantum, and it definitely fixes the issue.

@dan-palmer Merge, if you're happy?

dan-palmer:
👍

PR #237, opened 2015-07-31
Correct the order for 'bottom' leaderboard rankings

dan-palmer:
This ensures that the list returned from the `leaderboard`
endpoint in the `GeckoboardController` is presented in the 
correct order.

The head of the list will now yield the worst performer with
each consecutive element improving.

Prior to this fix the list was returned in descending
performance order(the worst performer would be the 10th element).


SteveMarshall:
:boom:

PR #240, opened 2015-08-04
Ensure replies aren't sent to postmaster@hmps

SteveMarshall:
6fb1abf didn't work, and failed to block postmaster@ emails because `p.from` is an instance of Mail::Address, which doesn't do string comparisons as we expected. Instead, we need to compare to
`p.from.address`.


SteveMarshall:
Merging myself in the absence of anyone else out of hours.

PR #241, opened 2015-08-04
Filtering out senders which are present on our blacklist.

lukaszraczylo:


app/controllers/webhooks_controller.rb
```diff
@@ -9,9 +9,11 @@ def email
         return
       end
 
-      if p.from.address == 'postmaster@hmps.gsi.gov.uk'
-        logger.error "Sender ( postmaster@hmps.gsi.gov.uk ) detected. Skipping message.\n #{p.inspect}"
-        render text: 'Sender postmaster@hmps.gsi.gov.uk'
+      blacklist = %w(postmaster sendmaster webmaster no-reply noreply root)
+
+      if blacklist.include? p.from.address.split('@')[0]
```
SteveMarshall: Instead of `p.from.address.split('@')[0]`, you should be able to use `p.from.local`, as in line 6 above (the `no-reply` check).

app/controllers/webhooks_controller.rb
```diff
@@ -9,9 +9,11 @@ def email
...
+        logger.error "Blacklisted sender detected. Skipping message.\n #{p.inspect}"
```
SteveMarshall: Do we want to dump the entire `inspect` here? If we get lots of mails, that might be noisy…

lukaszraczylo: Assuming that we received a bounce IMO it should be logged. "Proper" e-mails shouldn't give us any issues.

app/controllers/webhooks_controller.rb
```diff
@@ -9,9 +9,11 @@ def email
         return
       end
 
-      if p.from.address == 'postmaster@hmps.gsi.gov.uk'
-        logger.error "Sender ( postmaster@hmps.gsi.gov.uk ) detected. Skipping message.\n #{p.inspect}"
-        render text: 'Sender postmaster@hmps.gsi.gov.uk'
+      blacklist = %w(postmaster sendmaster webmaster no-reply noreply root)
```
SteveMarshall: Didn't @ashb mention that `root` shouldn't be included, or did you two decide it was fine to?

ashb: Yes - root is not a special local part, nor is webmaster for that matter.

app/controllers/webhooks_controller.rb
```diff
@@ -9,9 +9,11 @@ def email
...
+
+      if blacklist.include? p.from.local
```
ashb: We should lower-case the from address - technically PostMaster@a.com and postmaster@a.com are the same so lets just be safe about it.

PR #243, opened 2015-08-06
Re-enable anomalous prison slots on bank holidays

SteveMarshall:
In the move to the new PrisonSchedule system (3dcea63), we lost the ability to have anomalous prison slots on bank holidays.

To fix this, we only want to consider the holiday-ness of a day when evaluating if the day has “normal” visiting slots.

### DIFF
--- a/lib/prison_day.rb
+++ b/lib/prison_day.rb
@@ -6,7 +6,7 @@ class PrisonDay < Struct.new(:date, :prison)
   end
 
   def visiting_day?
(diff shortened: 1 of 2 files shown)


ashb:
This code works but the chain of "question functions" makes it a bit hard to follow the logic all the way down.

Maybe some comments or remove a layer or two of these might worth it to make it clearer?

SteveMarshall:
@dan-palmer @ashb's might be worth us having a look at when you're back, to see if we can clarify it a bit when you're back.

PR #245, opened 2015-08-10
Remove dead view for ID requirements

matharden:
This file is not being used `since standard_id_requirements` and `custom_id_requirements` [methods were removed](https://github.com/ministryofjustice/prison-visits/pull/214).


SteveMarshall:
Thanks @matharden! :fireworks:

PR #246, opened 2015-08-11
Find a prison rebased

matharden:
This is a rebased branch of `find-a-prison` and needs to replace it thus making it comparable to `master`. 

It is WIP. Requires tests and there is too much logic in the views.

Please take this and speak to @leighmoney about deploying somewhere so it can be viewed.


threedaymonk:
Closing stale PR as discussed.

SteveMarshall:
As discussed, this needs to stick around until I/we work out how to integrate it and serve it correctly.

kyliemulholland:
I reckon we either need to migrate this over to PVB2 or close it down. Thoughts @SteveMarshall ?

SteveMarshall:
@kyliemulholland Honestly, migrating it to the new application in its current form is going to be more effort than it's worth. We *should* create a little app that does this to replace the current prison finder, but it shouldn't be part of PVB.

kyliemulholland:
@SteveMarshall Cheers.

PR #247, opened 2015-08-12
Fix logstasher failures when receiving params

tatyree:
Logstasher would fail to log if any controller params came in encoded in
ASCII-8BIT.  I've created a drop-in gem to fix this and added it to the
gemfile.


threedaymonk:
This pull request is itself uncontroversial, but I've added some comments on the gem: https://github.com/ministryofjustice/string_scrubber/commit/de4e120e7d33546761a8f14f5623e4809a15ef23

threedaymonk:
Merged.

PR #249, opened 2015-08-13
Set confirmation date to be within number of lead days

dan-palmer:
APVU confirms the lead days range should be inclusive of
the current day(if its a working day at the prison).

For example, if a prison has a lead time of 3 days and a visitor
booked on a Tuesday then they should except a response within 3
days, the latest reply date being the Thursday.


matharden:
That's not what the majority of prisons have agreed. At least, that's not what's been happening for the last 12 months and that's not what's in [the documentation](https://github.com/ministryofjustice/prison-visits#response-times).

If you got ahead with this change, you need to notify each prison in advance and also update the [README](https://github.com/ministryofjustice/prison-visits/blob/master/README.md).

Alternatively you could set the `lead_days` configurations for the APVU establishments.

SteveMarshall:
@dan-palmer As @matharden says, the requirements for this are slightly different than I think you were told. Let's talk about it this morning.

SteveMarshall:
This is no longer necessary, thanks to #253. Closing.

PR #252, opened 2015-08-17
Sendgrid helper refactor

EGiataganas:


lib/sendgrid_helper.rb
```diff
@@ -1,18 +1,43 @@
...
+    end
+    true
+  rescue StandardError => e
+    false
+  end
+
+  private
+
+  def self.can_access_sendgrid?
+    ENV['SMTP_USERNAME'] and ENV['SMTP_PASSWORD']
```
SteveMarshall: Looking at this again, we should probably use the existing config variables, rather than directly accessing environment variables. I *think* you should be able to get them using `config.action_mailer.smtp_settings.user_name`?

config/environments/test.rb
```diff
@@ -1,6 +1,9 @@
 PrisonVisits2::Application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
+  ENV['SMTP_USERNAME'] = 'Test'
+  ENV['SMTP_PASSWORD'] = '12345678'
```
SteveMarshall: We probably shouldn't be directly setting (or using, as I mention below) environment variables. See if you can set `config.action_mailer.smtp_settings.user_name` and `config.action_mailer.smtp_settings.password`, maybe?

SteveMarshall:
It would be nice to see some tests for this change, something like “it calls sendgrid's API if the config variables exist”, and then “it doesn't talk to send grid if the config variables don't exist”.

SteveMarshall:
It would be nice the commit messages here better described the changes they contain, and represented clearer chunks. For example, 3133700, 598f9fd, and 9e65a3c could be combined into one commit called “Extract Sendgrid API query string generation”.

It helps to think of the first line of commit messages as completing the sentence “If applied, this commit will…”

config/environments/test.rb
```diff
@@ -1,6 +1,11 @@
 PrisonVisits2::Application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
+  config.action_mailer.smtp_settings = {
+    user_name: 'test_smtp_username',
+    password: 'test_smtp_password'
+  }
```
SteveMarshall: When running tests, they already mock these settings, so we shouldn't need this to be set, I don't think?

spec/lib/sendgrid_helper_spec.rb
```diff
@@ -2,62 +2,91 @@
...
-        SendgridHelper.spam_reported?('test@irrelevant.com')
+      context "when no error" do
+        it "marks an e-mail as valid" do
+          expect(JSON).to receive(:parse).and_return(['dummy'])
+          SendgridHelper.bounced?('test@irrelevant.com')
+        end
+
+        it "marks an e-mail as invalid" do
+          expect(JSON).to receive(:parse).and_return([])
+          SendgridHelper.bounced?('test@irrelevant.com')
```
SteveMarshall: I don't think these tests are actually evaluating truthiness of the return of `bounced?`, are they @threedaymonk?

threedaymonk: You're right. My bad: I didn't actually look at them when I indented them. They do not appear to test anything useful (or certainly not what they assert to test).

threedaymonk:
Tested on preprod.

spec/lib/sendgrid_helper_spec.rb
```diff
@@ -2,62 +2,116 @@
...
+      example.run
+      Rails.configuration.action_mailer.smtp_settings = smtp_settings
+    end
+
+    context "spam reports" do
+      context "error handling" do
+        [Curl::Err::CurlError, JSON::ParserError].each do |exception_class|
+          it "marks the email as valid if there's an error (#{exception_class})" do
+            allow(Curl::Easy).to receive(:perform).and_raise(exception_class)
+            expect(SendgridHelper.spam_reported?('test@example.com')).to be_falsey
```
dan-palmer: Two things:

##### 1.
Does `spam_reported?` always return instances of the `TrueClass` or `FalseClass`?

If so its better to show this in the assertion by using `be true` or `be false`, it reveals with less ambiguity the return value of the method. This applies to any other predicate methods that return only `true` or `false` values.

##### 2.
I'd suggest using RSpec's `described_class` when writing expectations about the current 'thing' in question. Its intended to return the class or module that is being described. If you aren't happy with the fact that it uses class in its name when talking about a `Module` then perhaps set it as the `subject` - but keep it consistent with the way other modules have been tested.

@threedaymonk @SteveMarshall do you guys agree?

threedaymonk: Regarding the second, to be honest, I'd rather rewrite this 'Helper' entirely to be (a) better tested and (b) not a module: I tried to make minimal changes here, and didn't rewrite the specs beyond some indentation and actually making them test what they claimed.

The first point is correct: it looks like they do always return a boolean. However, I suspect that having to use either `be true` or `be_truthy` is probably the fingerprint of a bad API, and it should be possible to do something like `expect(subject).not_to be_spam`.

PR #253, opened 2015-08-17
Fix calculation of confirmation date

threedaymonk:
Previously, the system calculated the confirmation date as the working day after the lead time. It should be the last day of the lead time.

For example, if a prison processes visits Monday to Friday, and Saturday is a visiting day, then it should be possible to book a visit for the coming Saturday:

    Tuesday   | Make booking
    Wednesday | Processing
    Thursday  | Processing
    Friday    | Processing & Confirmation
    Saturday  | Visit

Previously, because of the extra day, the confirmation would not be sent until Monday, and visits would not be possible until Tuesday.

See https://www.pivotaltracker.com/story/show/101219102

### DIFF
--- a/lib/prison_day.rb
+++ b/lib/prison_day.rb
@@ -2,11 +2,11 @@ class PrisonDay < Struct.new(:date, :prison)
   WEEKEND_DAYS = %w<sat sun>.freeze
 
   def staff_working_day?
-    non_holiday? && working_day?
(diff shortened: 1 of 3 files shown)


dan-palmer:
These are excellent improvements! 👍 Nice work @threedaymonk.

SteveMarshall:
This is great, @threedaymonk! :star:

PR #255, opened 2015-08-18
Add August bank holiday to Wymott

kyliemulholland:


kyliemulholland:
Have tested on PreProd and passed.

PR #256, opened 2015-08-19
Fix visit helper xss

EGiataganas:


spec/helpers/visit_helper_spec.rb
```diff
@@ -60,6 +60,17 @@ def visit
       expect(helper.prison_address).to start_with "1 Fort Road"
     end
 
+    context "with illegal html in the address" do
+      before :each do
+        expect(Rails.configuration.prison_data['Rochester']).to receive(:[]).
```
threedaymonk: This should probably be an `allow` as it's not the thing we're testing but is just configuration of the test environment. Personally, I'd say don't put expectations in `before` or `after` blocks and aim to have only one per spec (feature specs excepted).

spec/helpers/visit_helper_spec.rb
```diff
@@ -60,6 +60,17 @@ def visit
...
+        with('address') { ['Danger<script>ous'] }
+      end
+
+      it "doesn't return illegal html" do
+        expect(helper.prison_address).to eq("Danger")
```
threedaymonk: This is quite tightly coupled to the exact behaviour of `sanitize`. It might be better to test that it *doesn't* contain `<`.

SteveMarshall: :+1:

app/helpers/visit_helper.rb
```diff
@@ -46,7 +46,7 @@ def prison_slot_anomalies
   end
 
   def prison_address(glue='<br>')
-    prison_data['address'].join(glue).html_safe
+    prison_data['address'].map(&method(:sanitize)).join(glue).html_safe
```
threedaymonk: I don't really like to see `html_safe` called on user input (even if it has been sanitised), but there's a handy method that will already do just this: [safe_join](http://api.rubyonrails.org/classes/ActionView/Helpers/OutputSafetyHelper.html#method-i-safe_join).

app/helpers/visit_helper.rb
```diff
@@ -46,7 +46,7 @@ def prison_slot_anomalies
   end
 
   def prison_address(glue='<br>')
-    prison_data['address'].join(glue).html_safe
+    safe_join(prison_data['address'], glue.html_safe)
```
SteveMarshall: @threedaymonk I couldn't see another way than what @EGiataganas has done here to make the `glue` render as actual markup… Don't suppose you know of one?

spec/helpers/visit_helper_spec.rb
```diff
@@ -60,6 +60,14 @@ def visit
       expect(helper.prison_address).to start_with "1 Fort Road"
     end
 
+    it "provides illegal html in the address and doesn't return illegal html" do
+      allow(Rails.configuration.prison_data['Rochester']).
+        to receive(:[]).with('address') { ['Danger<script>ous', 'apoel'] }
```
SteveMarshall: Maybe change `apoel` in this to something else that's less… messy?

spec/helpers/visit_helper_spec.rb
```diff
@@ -60,6 +60,14 @@ def visit
...
+
+      expect(helper.prison_address).not_to match(/<script/)
+      expect(helper.prison_address).to match(/<br/)
```
SteveMarshall: @threedaymonk D'you reckon it's worth having two tests for this (one for single line, one for multi-line)? My gut says not, but I'm open to suggestion…

threedaymonk: I think I *would* split these up, but not for single line/multiple line: instead, I'd test *escapes HTML in the prison address* and *joins multiple lines with unescaped glue string*, which would end up with one expectation each.

SteveMarshall: Sounds right to me. @EGiataganas, can you do that?

app/helpers/visit_helper.rb
threedaymonk: No: even `content_tag` won't work (it will generate `<br></br>`). I would suggest a small change, though: to move the `.html_safe` to the default parameter, so that anyone overriding `glue` has to take responsibility for its safety.

SteveMarshall: :+1:

spec/helpers/visit_helper_spec.rb
EGiataganas: @SteveMarshall yeap

SteveMarshall:
Looks good to me! @threedaymonk?

threedaymonk:
Yep, looks good to me too.

Issue #225, opened 2015-06-25
Upgrade healthcheck.json to use same format as other services


SteveMarshall:
Other services in MOJ use this format:

```json
{
  "database": {
    "description": "Postgres database",
    "ok": true
  },
  "smtp": {
    "description": "SendGrid",
    "ok": true
  },
  "api": {
    "description": "DWP API",
    "ok": false
  },
  "ok": false
}
```

This provides a slightly more useful, descriptive set of information than our current format, so we should consider moving to it at some point soon.

threedaymonk:
Should we roll this into #258, as we're changing `healthcheck.json` anyway?

SteveMarshall:
Yeah, that'd be good, I think.

PR #258, opened 2015-08-20
Send email and tickets outside the main request-response cycle

threedaymonk:
**DO NOT MERGE YET** – Waiting for PIA

* https://www.pivotaltracker.com/story/show/96673886
* https://www.pivotaltracker.com/story/show/96673592

Sending email is one of the main performance-limiting factors in this application. This set of commits configures the application to use Redis for queueing, with the Sidekiq adapter. The queues are configured via `config/sidekiq.yml`, and the remaining configuration can be done via the command-line as required. The default configuration expects a local instance of Redis on its default port. The queue can be consumed using the `sidekiq` executable:

```sh
$ bundle exec sidekiq
```

Now that we are going to send emails via queues, we don't actually care about transient unavailability of SMTP servers or the Zendesk API. Instead, we care how many items are on the queues, and how old they are.

In order to facilitate monitoring of this, the healthcheck response now includes queue information, and we have taken this opportunity to make the format match that of other services as in #225. It now looks like this:

```json
{
  "database": {
    "description": "Postgres database",
    "ok": true
  },
  "mailers": {
    "description": "Email queue",
    "ok": true,
    "oldest": "2015-09-04T16:06:18.173Z",
    "count": 3
  },
  "zendesk": {
    "description": "Zendesk queue",
    "ok": true,
    "oldest": "2015-09-04T16:06:18.178Z",
    "count": 1
  },
  "ok": true
}
```

The mailers and zendesk checks pass if there are no stale items in those queues, which is computed by checking whether the oldest queue item is more than ten minutes old.

See also:

* [ActiveJob](http://edgeguides.rubyonrails.org/active_job_basics.html)
* [Available backends](http://api.rubyonrails.org/classes/ActiveJob/QueueAdapters.html)
* [Sidekiq Wiki](https://github.com/mperham/sidekiq/wiki)

### DIFF
--- a/app/services/healthcheck.rb
+++ b/app/services/healthcheck.rb
@@ -8,54 +8,54 @@ class Healthcheck
   end
 
   def ok?
-    checks.values.all?
+    checks.fetch(:ok)
   end
 
   def checks
-    {
-      database: database_active?,
-      mailers: fresh?('mailers'),
-      zendesk: fresh?('zendesk')
+    components = {
+      database: database,
+      mailers: mailers,
+      zendesk: zendesk
     }
-  end
-
-  def queues
-    {
-      mailers: queue_info('mailers'),
-      zendesk: queue_info('zendesk'),
-    }
+    components.merge(ok: components.values.map { |h| h.fetch(:ok) }.all?)
   end
 
   private
 
-  def queue(name)
-    @queues[name] ||= Sidekiq::Queue.new(name)
+  def database
+    {
+      description: 'Postgres database',
+      ok: database_active?
+    }
   end
 
-  def queue_info(queue_name)
-    q = queue(queue_name)
+  def mailers
+    queue_info('mailers', 'Email queue')
+  end
+
+  def zendesk
+    queue_info('zendesk', 'Zendesk queue')
+  end
+
+  def queue_info(name, description)
+    q = Sidekiq::Queue.new(name)
     {
-      oldest: oldest_item_created_at(q),
+      description: description,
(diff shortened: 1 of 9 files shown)


SteveMarshall:
This is looking great :+1:

app/services/healthcheck.rb
```diff
@@ -0,0 +1,60 @@
+require 'sidekiq/api'
+
+class Healthcheck
```
SteveMarshall: These should probably all `rescue`, just in case something goes wrong, so we can give some sort of default healthcheck output even if everything downstream is broken? (We just had exactly this on preprod.)

threedaymonk: That sounds like a good idea. Were you thinking of something like this:

```ruby
def checks
  {
    # ...
   }
rescue
  { all: 'broken' }
end
```

or trapping each of the key methods (e.g. `database_active?`; `old_items?` might need its logic reversed) to return `false` on error?

SteveMarshall: I'd err toward the latter, mostly because at the moment it seems to fail entirely if any one of the checks breaks, so it might be a false positive to say `all: 'broken'`.

threedaymonk: Done.

SteveMarshall:
@threedaymonk PIA is approved, apparently. Can you rebase this against master so we can test on preprod?

SteveMarshall:
Tested this on preprod. Let's :shipit: tomorrow morning!

PR #260, opened 2015-08-21
Fix CSRF protection at the ApplicationController

EGiataganas:
Fix the implementation of CSFR protection at thet Application Level by
setting `protect_from_forgery with: :exception` in the
`ApplicationController`

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,7 @@
 class ApplicationController < ActionController::Base
   helper_method :visit
   before_filter :add_extra_sentry_metadata
+  protect_from_forgery with: :exception
 
   def self.permit_only_from_prisons
     before_filter :reject_untrusted_ips!
--- a/app/controllers/feedbacks_controller.rb
+++ b/app/controllers/feedbacks_controller.rb
@@ -1,5 +1,4 @@
 class FeedbacksController < ApplicationController
-  protect_from_forgery
 
   def index
   end


EGiataganas:
@SteveMarshall Tested on preprod!

dan-palmer:
This is fine but it would be nice if we could test it. 

Its kind of fiddly to test through a feature test but perhaps we could do something with the `forgery_protection_strategy` config value that `ActionController::Base` has. Calling this should return whatever the strategy is if it has been set in the controller or nil. Perhaps we could have a shared example that all the controller tests have that asserts forgery protection is enabled. Its quite implementation specific doing it this way but I can't think of an alternate thats cleaner off the top of my head.

Thoughts @SteveMarshall, @threedaymonk ?

SteveMarshall:
@dan-palmer I'm not sure that would work: we disable forgery protection in the tests, so any tests checking it would find it disabled and, because it's set in config, it's set (unless I'm mistaken?) for the lifetime of the rails server (and, by extension, the lifetime of the test run?), so we can't change it (as @EGiataganas and I discovered yesterday) mid-test.

I'd like to test this, but I'm not sure I can see a clear way to do that.

threedaymonk:
Personally, I'm not convinced of the necessity of testing code that is merely declarative, and as this is set in the top-level controller from which all others derive, I don't think there's a lot of value in testing each controller.

I'm happy to take it on trust that Rails works as designed – or, at least, that CSRF protection is tested in more detail in Rails than we'd ever do in our application's suite.

However, the tests run just fine with

```ruby
config.action_controller.allow_forgery_protection = true
```

so I think the option is there if you want it.

EGiataganas:
@threedaymonk Yesterday @SteveMarshall and I we were trying to fake the 
```
config.action_controller.allow_forgery_protection = true
```
by making that <a href="https://gist.github.com/EGiataganas/2141370e2457e157b699">gist</a> using around. But we could find 
```
<input type="hidden" name="authenticity_token" value="">
```
so we couldn't test it. I'm happy too, to take it on trust that Rails works as designed – or, at least, that CSRF protection is tested in more detail in Rails than we'd ever do in our application's suite.

dan-palmer:
:ship:

threedaymonk:
I'm happy to merge this.

PR #261, opened 2015-08-24
Highpoint south 31st august

kyliemulholland:


SteveMarshall:
Looks like this was created from an old branch in a weird state. Can you recreate again from master?

PR #262, opened 2015-08-24
High down training dates

kyliemulholland:


kyliemulholland:
Have tested on PreProd - passed

PR #263, opened 2015-08-25
Add Zendesk to the healthcheck JSON endpoint

threedaymonk:
* https://www.pivotaltracker.com/story/show/96673886 (second check list item)


spec/controllers/healthcheck_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+    end
+
+    it 'returns an HTTP Bad Gateway status' do
+      expect(response).to have_http_status(:bad_gateway)
+    end
+
+    it "reports #{service} as inaccessible" do
+      expect(parsed_body).to include('checks' => include(service => false))
+    end
+  end
```
SteveMarshall: This is a really neat use of `shared_examples` :+1:

SteveMarshall:
Looks good to me :stars: @dan-palmer?

app/controllers/healthcheck_controller.rb
```diff
@@ -26,4 +21,14 @@ def sendgrid_alive?
   def messagelabs_alive?
     SendgridHelper.smtp_alive?(*PrisonMailer.smtp_settings.values_at(:address, :port))
   end
+
+  def database_active?
+    ActiveRecord::Base.connection.active?
```
dan-palmer: I think this is the only healthcheck checking the DB that works across all of the apps in MOJ as `ActiveRecord::QueryCache` middleware has been <a href='https://github.com/ministryofjustice/prison-visits/blob/master/config/application.rb#L73'>removed</a>.

spec/controllers/healthcheck_controller_spec.rb
threedaymonk: They were there before! I just pulled them out of the support file they were hidden in.

dan-palmer:
Looks good to me @threedaymonk, @SteveMarshall :+1:

spec/controllers/healthcheck_controller_spec.rb
SteveMarshall: Apparently I wrote them! Yay for memory! :dizzy_face:

app/controllers/healthcheck_controller.rb
```diff
@@ -26,4 +21,14 @@ def sendgrid_alive?
   def messagelabs_alive?
     SendgridHelper.smtp_alive?(*PrisonMailer.smtp_settings.values_at(:address, :port))
```
dan-palmer: Not related to @threedaymonk's work in this PR.. I don't like the fact that what should probably be `MailerHelper` or similar is called `SendgridHelper`.

Knowing that the `PrisonMailer` isn't using Sendgrid is a little confusing. :laughing:

threedaymonk: I agree! I'm also not a fan of most of these 'Helper' module function things that litter the codebase.

threedaymonk:
Tested on preprod.

PR #266, opened 2015-08-25
Include name of not-found prison in error message

threedaymonk:
This will assist in tracking down the reason behind 500 errors caused by `PrisonNotFound` exceptions.


dan-palmer:
@threedaymonk, @SteveMarshall  - my bad on this one, I should have included it when I created the model. 

However the reason for the 500's is because of an empty visit object in the session(redirects on session timeouts is the cause of this) so don't be alarmed when you just see `Can't find prison ` in the log file/kibana.

PR #267, opened 2015-08-27
Add August bank holiday as available for Buckley hall

kyliemulholland:


kyliemulholland:
Can't test on PreProd as 31st Aug is less that 3 working days from today :(

SteveMarshall:
D'oh! Well, not much point merging this, then…

PR #272, opened 2015-08-28
Tests and fixes for unexpected journeys

threedaymonk:
This PR adds specs and fixes for journeys that are outside the expected path, i.e. where the session expires, or where the back button is used.

This seems to rule out session expiry as a cause of 500 errors, but does identify a large class of problems that can happen when a visitor jumps straight to a later page without the expected information in their session.

After the first step, the session must contain a visit, which has a prisoner, who has a prison name. Without this, the prison data cannot be retrieved and an exception results.

Now, if the session doesn't contain this information, we'll redirect the visitor back to the first step and get them to start again. This won't solve every possible 500 error, but it should make a difference.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -72,4 +72,11 @@ class ApplicationController < ActionController::Base
   def metrics_logger
     METRICS_LOGGER
   end
+
+  def ensure_visit_integrity
+    unless visit && visit.prisoner && visit.prisoner.prison_name.present?
+      redirect_to edit_prisoner_details_path,
+        notice: 'You need to complete missing information to start or continue your visit request'
+    end
+  end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -36,4 +36,5 @@ group :test, :development do
   gem 'rspec-html-matchers'
   gem 'simplecov', '~> 0.7.1', require: false
   gem 'simplecov-rcov', require: false
+  gem 'launchy'
 end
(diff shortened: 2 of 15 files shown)


threedaymonk:
Tested and working on preprod.

app/controllers/application_controller.rb
```diff
@@ -72,4 +72,11 @@ def encryptor
   def metrics_logger
     METRICS_LOGGER
   end
+
+  def ensure_visit_integrity
```
dan-palmer: Cool  :sunglasses:

app/controllers/application_controller.rb
```diff
@@ -72,4 +72,11 @@ def encryptor
...
+    unless visit && visit.prisoner && visit.prisoner.prison_name.present?
+      redirect_to edit_prisoner_details_path,
+        notice: 'We’re missing some vital information'
```
SteveMarshall: This is great! @leighmoney do you want to adjust the copy at all for this?

leighmoney: Thanks for running by me @SteveMarshall. Is this like an error? Suggest we should say something like: 
++ 
You need to enter the prisoner's name

dan-palmer:
This is great stuff, nice work @threedaymonk.

app/controllers/application_controller.rb
threedaymonk: @leighmoney This is the message that they'll see if they go directly to the middle of the process, either by going to the URL directly or by doing something like clicking back after a session timeout. Previously, the app just blew up; now, they're sheep-dipped through the front gate again. It's not the name we need, as such, but *all* the information!

leighmoney: Ah Ok @threedaymonk Maybe: 

++
You need to complete missing information to start or continue your visit request
++ 

Hopefully that makes more sense?

PR #276, opened 2015-09-02
Let visitors override spam/bounced email rejection

threedaymonk:
This improves the reporting of email errors when an address has been reported by Sendgrid as having bounced or been marked as spam.

It adds a checkbox to the visitor form to allow the visitor to force acceptance of an email address that failed the Sendgrid checks. The checkbox is only displayed if validation fails for one of these two reasons. If selected, we permit the email address, and tell Sendgrid to remove the spam or bounced categorisation before sending the email.

### DIFF
--- a/.default_rubocop.yml
+++ b/.default_rubocop.yml
@@ -66,6 +66,11 @@ Style/StringLiterals:
   Exclude:
     - 'spec/**/*'
 
+# Allow us to define simple predicate methods like foo? that look at the value
+# of @foo. See https://github.com/bbatsov/rubocop/issues/2140
+Style/TrivialAccessors:
+  AllowPredicates: true
+
 # Prefer sensible naming to comments everywhere.
 Documentation:
   Description: Document classes and non-namespace modules.
--- a/app/mailers/visitor_mailer.rb
+++ b/app/mailers/visitor_mailer.rb
@@ -41,6 +41,8 @@ class VisitorMailer < ActionMailer::Base
     @visit = visit
     @token = token
 
+    perform_sendgrid_resets
+
     mail(from: sender,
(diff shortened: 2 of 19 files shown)


SteveMarshall:
This is looking great so far! :sunglasses:

threedaymonk:
Design:

email - spam

Reality:

visitor_1_-_visit_someone_in_prison_-_2015-09-03_12 08 15

The bottom half is OK (it follows the design, modulo changes in template that should be addressed elsewhere). However, the fact that error messages are repeated at the top and bottom in different contexts is problematic; in fact, it's a problem generally, and not just here.

SteveMarshall:
@threedaymonk That's mostly looking good. As for the messaging context issue… I think that's another one for @leighmoney or @ed209uk?

tonyduarte:
@SteveMarshall @threedaymonk...  or me ;)

will look at it with eddie (@addie not showing for me)

threedaymonk:
@tonyduarte The messages that it's expecting for the top box are generally in the form of sentence fragments like this: [Rails's defaults](https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/locale/en.yml) or [in this app](https://github.com/ministryofjustice/prison-visits/blob/8b7a5b49480505e6c8018143a440fc10637e558c/config/locales/en.yml#L35-L50). Unfortunately, the validation framework only lets us have **one** kind of message per field, so it's very difficult to show different messages in different contexts.

threedaymonk:
Tested the bounce override with an email address I previously caused to bounce:

screenshot-2015-10-02t10 54 19z

screenshot-2015-10-02t10 54 40z

PR #275, opened 2015-09-02
Rochester add Friday slot

kyliemulholland:
Waiting for Jenkins to come alive before I can test in preprod...

PR #277, opened 2015-09-03
Decouple Rochester slots from visit_helper_spec

SteveMarshall:
We don't want to have to update our tests every time Rochester's visit slots change, so mock them.

Also reorder the expected slots so they match how weeks actually work.

This will fix the issues with #275.

### DIFF
--- a/spec/helpers/visit_helper_spec.rb
+++ b/spec/helpers/visit_helper_spec.rb
@@ -22,13 +22,28 @@ RSpec.describe VisitHelper, type: :helper do
     end
 
     it "provides a hash of slots by day" do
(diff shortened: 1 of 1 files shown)


threedaymonk:
On consideration, I wonder whether it might be better to define a hash with *all* the canned prison data at the top of the file, so that it's insulated against all future data changes:

```ruby
before do
  allow(Rails.configuration).to receive(:prison_data).and_return(
    'Rochester' => {
      # etc.
    }
  )
end
```

Better still might be to define the relevant subset in each example, but I don't think it's worth investing too much in this spec given that the whole thing is flawed: the best of all would be for this 'helper' to go away entirely and for this information to be available by interrogating the Prison object.

PR #278, opened 2015-09-03
Thorn cross add saturday slot

kyliemulholland:
Tested in preprod - passed.


SteveMarshall:
@kyliemulholland Merged. Deploy away!

PR #279, opened 2015-09-04
Prepare to queue email and tickets

threedaymonk:
* https://www.pivotaltracker.com/story/show/96673886
* https://www.pivotaltracker.com/story/show/96673592

Sending email is one of the main performance-limiting factors in this application. This set of commits implements queueing with ActiveJob, and changes `deliver_now` to `deliver_later` and makes the necessary changes for that to work. It also moves Zendesk ticket submission to a job using `perform_later`.

However, actually deploying queueing requires both server configuration and the PIA to be signed off. In order to make our deployments smaller, this PR makes the preparatory steps needed to implement queueing, but uses Rails's default `InlineAdapter`, which executes jobs immediately in the same process.

This means that the changes deployed in #258 will be much smaller and more focused.

See also:

* [ActiveJob](http://edgeguides.rubyonrails.org/active_job_basics.html)
* [Available backends](http://api.rubyonrails.org/classes/ActiveJob/QueueAdapters.html)


threedaymonk:
Tested on preprod. Email and Zendesk confirmed working.

SteveMarshall:
:boom:

PR #280, opened 2015-09-04
Add sidekiq web admin UI rackup script and required gems

ashb:
Sidekiq has a nice admin interface that gives us queue status, and can review errors etc.

Rather than using basic auth we are going to put this behind an OAuth2 gateway (run separately) and configured in the load-balancer to live under /sidekiq.

Although sidekiq is sinatra and could be mounted in the PVB app putting just one path behind OAuth would involve installing omniauth and we felt this was a smaller and safer change to make it happen this way. (It also means that there's less scope that a mistake in the config/routes.rb could open it up to the world, but that was not the driving factor behind doing it this way.)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,7 @@ gem 'statsd-ruby', require: 'statsd'
 gem 'curb'
 gem 'string_scrubber', '>= 0.2.0'
 gem 'virtus'
+gem 'sinatra'
 
 group :test, :development do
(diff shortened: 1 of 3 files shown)


threedaymonk:
This should be rebased onto #279.

SteveMarshall:
@pdb … and now master.

SteveMarshall:
Closing because this needs to be merged into master now.

PR #282, opened 2015-09-04
Update Belmarsh visit slots and email addresses

kyliemulholland:


kyliemulholland:
@SteveMarshall Let me know if I need to fix something.

SteveMarshall:
@kyliemulholland The error that's occurring is because the unbookable dates differ. That said, I'll rework it because there's a few conflicts with master…

SteveMarshall:
@kyliemulholland Updated and fixed. Wanna try on preprod?

PR #283, opened 2015-09-08
Belmarsh amend booking window

kyliemulholland:


kyliemulholland:
Tested in preprod - passed.

PR #284, opened 2015-09-08
Belmarsh add saturday morning slot

kyliemulholland:


kyliemulholland:
Tested in preprod - passed.

PR #285, opened 2015-09-08
Improve visitor model and specs

threedaymonk:
This makes various improvements to the way that the visitor model and validations work and are tested, and corrects age calculation (as this was flagged by the new tests). It is intended to lay the groundwork for #276.


threedaymonk:
Tested on preprod.

SteveMarshall:
LGTM!

app/services/age_calculator.rb
```diff
@@ -0,0 +1,12 @@
+class AgeCalculator
+  def age(date_of_birth, today)
+    # See https://stackoverflow.com/a/2357790
+    if today.month > date_of_birth.month ||
```
dan-palmer: How about hoisting this guy into a predicate method, something like `had_bithday_this_year?`

spec/features/datepicker_spec.rb
```diff
@@ -15,9 +15,6 @@
 
   before :each do
     allow_any_instance_of(VisitController).to receive(:metrics_logger).and_return(MockMetricsLogger.new)
-    allow_any_instance_of(EmailValidator).to receive(:validate_dns_records)
```
dan-palmer: :door: :walking:  :sunglasses: nice one!

app/services/age_calculator.rb
threedaymonk: Good call.

threedaymonk:
Tested on preprod (as at ff0823c).

dan-palmer:
LGTM! 😎

PR #286, opened 2015-09-09
Revert "Improve visitor model and specs"

SteveMarshall:
Reverts ministryofjustice/prison-visits#285 because I'm an idiot and clicked too soon.


SteveMarshall:
Don't need this. I force-pushed to master. Still an idiot, though.

PR #288, opened 2015-09-09
Kennet update visit slots

kyliemulholland:


kyliemulholland:
Tested on preprod - passed.

PR #289, opened 2015-09-10
Report Coverage to Code Climate

EGiataganas:
Did the basic installation for code climate. Added the gem file and then
added test reporter on the very first line of the spec_helper.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -76,6 +76,8 @@ GEM
       nokogiri (~> 1.6)
       version_sorter (~> 1.1)
     cliver (0.3.2)
+    codeclimate-test-reporter (0.4.8)
+      simplecov (>= 0.7.1, < 1.0.0)
     coderay (1.1.0)
     coercible (1.0.0)
       descendants_tracker (~> 0.0.1)
@@ -314,6 +316,7 @@ PLATFORMS
 DEPENDENCIES
   capybara
   chromedriver-helper
+  codeclimate-test-reporter
   coffee-rails
   curb
   database_cleaner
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -1,3 +1,6 @@
+require "codeclimate-test-reporter"
+CodeClimate::TestReporter.start
+
 require 'simplecov'
 require 'simplecov-rcov'
 SimpleCov.formatter = SimpleCov::Formatter::RcovFormatter
```
threedaymonk: I think that we want to run simplecov only if the CodeClimate environment variable is *not* set, so this should be something like:

```ruby
if ENV.key?('CODE_CLIMATE_WHATEVER')
  require "codeclimate-test-reporter"
  CodeClimate::TestReporter.start
else
  require 'simplecov'
  require 'simplecov-rcov'
  SimpleCov.formatter = SimpleCov::Formatter::RcovFormatter
end
```

dan-palmer:
Nice to see you using `fixup` :star2:

threedaymonk:
There's an element of uncertainty here in that we're not sure whether it will actually report coverage until it's merged into the master branch, because CodeClimate will only report coverage on the master branch, and it won't be reporting on the same commit until then.

Still, I think this is fine to merge.

Gemfile.lock
```diff
@@ -350,3 +353,6 @@ DEPENDENCIES
   unicorn
   virtus
   zendesk_api
+
+BUNDLED WITH
+   1.10.6
```
threedaymonk: @dan-palmer @SteveMarshall Shall we bite this bullet?

dan-palmer:
Can confirm this as working :ship:

PR #293, opened 2015-09-14
Add temporary Money to Prisoners survey to request send page

teneightfive:
This adds a temporary survey to gather more feedback for the Money to Prisoners service. 

It is essentially a duplication of PR #138 but with a new [survey](https://www.surveymonkey.com/r/PJWKMSQ).

### DIFF
--- a/app/assets/javascripts/shame.js
+++ b/app/assets/javascripts/shame.js
@@ -45,6 +45,11 @@ $(function () {
 
     $(this).closest('.BookingCalendar').siblings('p').find('span').text(days[date.getDay()] +' '+ date.getDate() +' '+ months[date.getMonth()]);
   });
+
+  // Temporary survey
+  $('#survey-no-thanks').on('click', function() {
+    $('#user-satisfaction-survey-container').remove();
+  });
 });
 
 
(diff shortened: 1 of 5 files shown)


app/views/layouts/application.html.haml
```diff
@@ -47,6 +47,10 @@
   = succeed '.' do
     = link_to "Find out more about cookies", cookies_path
 
+- content_for :after_header do
```
dan-palmer: shouldn't we just put this in: https://github.com/ministryofjustice/prison-visits/blob/master/app/views/deferred/visits/show.html.haml

teneightfive: Happy to make that change if it's only used for that view and not shared anywhere else. I was copying the original changes from #138.

teneightfive: Can a view template contain a `content_for` block actually? Not sure if they need to be in `application.html.haml` and loaded before the `= render template` line.

dan-palmer:
screen shot 2015-09-14 at 15 55 27

@teneightfive LGTM :+1:  Is it as expected?

PR #294, opened 2015-09-14
Use a third-party Sendgrid API library

threedaymonk:
There exists a gem for integrating with Sendgrid's API: [sendgrid_toolkit](https://github.com/freerobby/sendgrid_toolkit). So let's not rewrite it.

For the existing API integration, this PR represents just a slight improvement in terseness, but it provides a good basis for adding new actions.

`SendgridHelper` is now `SendgridApi`, a thin layer that preserves the existing interface whilst using `sendgrid_toolkit` underneath. This is a minimal change in terms of testing and interface, but we should now be confident enough that we only test that we're using the `sendgrid_toolkit` API correctly in future.

### DIFF
--- a/app/services/sendgrid_api.rb
+++ b/app/services/sendgrid_api.rb
@@ -6,14 +6,14 @@ class SendgridApi
   def spam_reported?(email)
     return false unless can_access_sendgrid?
     spam_reports.retrieve(email: email).any?
-  rescue SendgridToolkit::APIError
+  rescue JSON::ParserError, SendgridToolkit::APIError
     false
   end
 
   def bounced?(email)
     return false unless can_access_sendgrid?
     bounces.retrieve(email: email).any?
-  rescue SendgridToolkit::APIError
+  rescue JSON::ParserError, SendgridToolkit::APIError
     false
   end
 
--- a/app/services/sendgrid_api.rb
+++ b/app/services/sendgrid_api.rb
@@ -1,46 +1,56 @@
-module SendgridHelper
-  def self.spam_reported?(email)
+class SendgridApi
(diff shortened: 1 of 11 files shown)


app/services/sendgrid_api.rb
```diff
@@ -0,0 +1,56 @@
+class SendgridApi
+  extend SingleForwardable
+
+  def_single_delegators :new, :spam_reported?, :bounced?, :smtp_alive?
+
+  def spam_reported?(email)
```
dan-palmer: :thought_balloon: Perhaps we could DRY the predicate methods up a bit.. however I am wondering if these two methods should be classed as just coincidentally similar for the time being and it might be a bit naive to extract a generalised helper method. If not perhaps something like this:

```ruby
class SendgridApi
  def spam_reported?(email)
    such_api { spam_reports.retrieve(email: email).any? }
  end

  def bounced?(email)
    such_api { bounces.retrieve(email: email).any? }
  end

  private

  def such_api
    return false unless can_access_sendgrid?
    yield
  rescue JSON::ParserError, SendgridToolkit::APIError
    false
  end
end
```

dan-palmer:
LGTM.. nice one :+1:

app/services/sendgrid_api.rb
threedaymonk: There definitely *is* duplication. I didn't think it was worth trying to extract any commonality right now, because I'm expecting to add a couple more functions (to cancel bounces and spam reports) that will have slightly different behaviour.

dan-palmer: Totes cool :+1: :ship:

PR #268, opened 2015-08-27
Add smoke test for PVB user journey

dan-palmer:
Adds a smoke test flow to the application - found under `smoke_test/`.

Adds a `PrisonMailer` check for smoke tests - if its a smoke test avoid sending stuff to the actual prison!

##### Smoke Test Steps:
  - [x] Starts the form
  - [x] Fills in prisoner details page
  - [x] Fills in visitor details page for one visitor
  - [x] Selects three available days on the slot page
  - [x] Submits the booking on the check your request page
  - [x] Checks that the user received a booking receipt email
  - [x] Checks that the prison received a booking request email
  - [x] Visits the process booking page and confirms the booking
  - [x] Checks that the prison received a copy of the confirmation email
  - [x] Checks that the visitor received a confirmation email
  - [x] Checks the status of the visit and cancels it
  - [x] Checks that the prison receives a booking cancellation email

#### Improvements
- [x] Handle sending Prison mails through Sendgrid(see: #297)
- [x] Use unique gmail addresses (this one cool '+' trick)
- [x] Auto run the smoke test
- [x] Separate concerns a wee bit
- [x] Start the process form the '/prisoner' page to avoid routing issues with production

##### Required environment variables for the rails app:
```
SMOKE_TEST_EMAIL_LOCAL_PART
SMOKE_TEST_EMAIL_DOMAIN
```

##### Required environment variables for the smoke test:
```
SMOKE_TEST_EMAIL_LOCAL_PART
SMOKE_TEST_EMAIL_DOMAIN
SMOKE_TEST_EMAIL_PASSWORD
SMOKE_TEST_APP_HOST
```

PR #297, opened 2015-09-15
Ensure "quoted-printable" email encoding

dan-palmer:
Sendgrid enforces line length limits of 998 characters. This is according to the <a href="http://tools.ietf.org/html/rfc5322">RFC spec</a> see section 2.1.1 for more info.

We have noticed that ActionMailer has some inconsistencies with what Content-Transfer-Type header is set for Prisoner & Visitor emails. The VisitorMailer emails are nearly always set as 'quoted-printable' whereas the Prison emails are '7bit'. Sending through Gmail & GSI does not alter the contents of the email whereas Sendgrid will enforce the rulings of the RFC spec unless the transfer encoding is set to 'quoted-printable' or 'base64' and add line breaks which will break long links as found in the PrisonMailer emails.

This addition adds a filter than can be mixed into any of the mailers we have in the application and will ensure the Content-Transfer-Type header is always set to 'quoted printable'

Quoted printable is an encoding that allows us to transmit 8 bit data over a 7 bit medium and limits line length to 76 characters by using a '=' as a separator. It produces human readable output that is rendered correctly by the likes of Gmail for both plain text and html emails.

This issue was brought to our attention as the smoke tests found in PR #268 override the smtp settings with the default action mailer settings which use Sendgrid - the emails had broken links.

This is a shameful hack.


app/mailers/concerns/addresses.rb
```diff
@@ -0,0 +1,15 @@
+module Addresses
+  include ActiveSupport::Concern
+
+  def prison_mailbox_email
+    Rails.configuration.prison_data[@visit.prisoner.prison_name]['email']
```
threedaymonk: Just imagine: one day, we can change this to something sensible like `@visit.prison_email`! But maybe not today.

dan-palmer: 😂🙌

SteveMarshall:
Bit late to the party, but :100: for investigation, @dan-palmer!

PR #298, opened 2015-09-16
Fix assignment in condition

EGiataganas:


app/controllers/application_controller.rb
```diff
@@ -37,15 +37,15 @@ def reject_without_key!
...
-      Raven.extra_context(visit_id: visit.visit_id)
-      if prison = visit.prisoner.prison_name
-        Raven.extra_context(prison: prison)
-      end
-    end
+    {
+      request_id: request_id,
+      visit_id: visit_id,
+      prison: prison_name
+    }.map { |key, value| Raven.extra_context(key => value) if value }
```
threedaymonk: I don't know what Raven is, but it looks like it has side effects. If so, perhaps `each do … end` might be more intention-revealing

app/controllers/feedbacks_controller.rb
```diff
@@ -9,9 +9,8 @@ def new
 
   def create
     @feedback = Feedback.new(feedback_params)
-    if visit = session[:visit]
-      @feedback.prison = visit.prisoner.prison_name
-    end
+    @feedback.prison = prison_name if visit
```
threedaymonk: I think you've accidentally dropped the assignment here, so `visit` will always be `nil`.

config/initializers/logstasher.rb
```diff
@@ -1,8 +1,9 @@
 if LogStasher.enabled
   LogStasher.add_custom_fields_to_request_context do |fields|
-    if visit = session[:visit] || session[:booked_visit]
+    visit = session[:visit]
+    if visit || session[:booked_visit]
       fields[:visit_id] = visit.visit_id
-      fields[:prison] = visit.prisoner.prison_name
+      fields[:prison] = visit.prison_name
```
threedaymonk: :+1: Good use of `delegate`

config/initializers/logstasher.rb
```diff
@@ -1,8 +1,9 @@
 if LogStasher.enabled
   LogStasher.add_custom_fields_to_request_context do |fields|
-    if visit = session[:visit] || session[:booked_visit]
+    visit = session[:visit]
+    if visit || session[:booked_visit]
```
threedaymonk: I think this needs to be:

```ruby
visit = session[:visit] || session[:booked_visit]
if visit
  …
```

config/initializers/session_store.rb
```diff
@@ -1,4 +1,5 @@
-SERVICE_DOMAIN = (url = ENV['SERVICE_URL']) ? URI.parse(url).host : nil
+SERVICE_URL = ENV['SERVICE_URL']
+SERVICE_DOMAIN = (SERVICE_URL) ? URI.parse(SERVICE_URL).host : nil
```
threedaymonk: If `SERVICE_URL` isn't used anywhere else, keep it as a variable (`service_url`) instead of a constant.

lib/google_analytics_adapter.rb
```diff
@@ -3,20 +3,38 @@ def initialize(visit)
...
   end
 
   def number_of_slots
-    nonempty_slots.count
+    date_sorted_slots.count
   end
 
   def slot_times
-    nonempty_slots.map { |s| s.times }
+    date_sorted_slots.map { |s| s.times }
```
threedaymonk: Could this be `date_sorted_slots.map(&:times)`?

lib/google_analytics_adapter.rb
```diff
@@ -3,20 +3,38 @@ def initialize(visit)
...
   end
 
   def slot_weekdays
-    nonempty_slots.map { |s| s.weekday }
+    date_sorted_slots.map { |s| s.weekday }
```
threedaymonk: As above, how about `date_sorted_slots.map(&:weekday)`?

config/initializers/logstasher.rb
EGiataganas: :+1: I agree, I will change it now!

lib/tasks/browserstack.rake
```diff
@@ -24,9 +24,9 @@ namespace :browserstack do
       # Fire up a tunnel in the background and wait until it is ready.
       r, w = IO.pipe
       pid = spawn("java -jar vendor/BrowserStackTunnel.jar -skipCheck #{ENV['BS_PASSWORD']} localhost,3000,0", out: w)
-      while content = r.readline
-        break if content == "Press Ctrl-C to exit\n"
-      end
+      content = r.readline
+
+      break if content == "Press Ctrl-C to exit\n" while content
```
threedaymonk: I don't think this works the same: in the old form, `r.readline` is evaluated on every loop.

app/controllers/application_controller.rb
EGiataganas: Ok, I'm happy to change it! :)

app/controllers/feedbacks_controller.rb
EGiataganas: Actually not, the `visit` comes from [[here](https://github.com/ministryofjustice/prison-visits/blob/master/app/controllers/application_controller.rb#L64)]

lib/tasks/browserstack.rake
EGiataganas: Yeah, I think you re right. What do you suggest?

lib/google_analytics_adapter.rb
EGiataganas: We actually did it, but then we thought is not a change due to rubocop so we were thinking to create a new branch and add those changes!

lib/google_analytics_adapter.rb
EGiataganas: We actually did it, but then we thought is not a change due to rubocop so we were thinking to create a new branch and add those changes!

lib/tasks/browserstack.rake
threedaymonk: I think this will work:

```ruby
loop do
  content = r.readline
  break if content.nil? || content == "Press Ctrl-C to exit\n"
end
```

lib/tasks/browserstack.rake
```diff
@@ -24,8 +24,11 @@ namespace :browserstack do
       # Fire up a tunnel in the background and wait until it is ready.
       r, w = IO.pipe
       pid = spawn("java -jar vendor/BrowserStackTunnel.jar -skipCheck #{ENV['BS_PASSWORD']} localhost,3000,0", out: w)
-      while content = r.readline
-        break if content == "Press Ctrl-C to exit\n"
+      content = r.readline
+
+      loop do
+        content = r.readline
+        break if content.nil? || content == "Press Ctrl-C to exit\n"
```
dan-palmer: cool!

lib/visitors_manipulator.rb
```diff
@@ -11,7 +11,8 @@ def adjust_visitors_in_session
   end
 
   def update
-    if m = params[:next].match(/remove-(\d)/)
+    m = params[:next].match(/remove-(\d)/)
```
dan-palmer: what is m? :last_quarter_moon_with_face:

dan-palmer: Upon closer inspection of this code it appears as though it won't ever get executed. As far as I can see `params[:next]` will always just be the value of the button which is either 'Continue' or 'Add another visitor'. This will apply for both JS and non-JS page renders. Assume its intent is to remove visitors that have been added however the `build_visitors` method in the module will handle this for us as its just takes all the visitor data from the params and creates new models unless the `_destroy` value has been set in the form. I'd suggest investigating this outside of this PR as its not in scope. Test will fail if its removed, assume thats because the tests know too much about the implementation though. Also suggest refactoring it a wee bit as it does all the things.

cc @threedaymonk @SteveMarshall @kyliemulholland

dan-palmer:
Suggest looking at the `update` method in the `VisitorsManipulator` module I mentioned ^^ as a separate task.

dan-palmer:
LGTM :+1:

PR #310, opened 2015-09-18
Fix unused method argument

EGiataganas:


lib/leaderboard_report.rb
```diff
@@ -60,7 +60,7 @@ def initialize(results)
       @results = results
     end
 
-    def as_json(options={})
```
dan-palmer: Why does this even take arguments? :hushed:

dan-palmer:
@EGiataganas has conflicts :boom:

PR #312, opened 2015-09-21
Wandsworth remove 0830-0930 visit slots

kyliemulholland:


kyliemulholland:
Tested on preprod - passed.

PR #314, opened 2015-09-21
Update Wandsworth email address to APVU

kyliemulholland:
DO NOT MERGE until 9am Tuesday 22nd Sept

### DIFF
--- a/config/prison_data_production.yml
+++ b/config/prison_data_production.yml
@@ -2678,7 +2678,7 @@ Wandsworth:
   canned_responses: true
   enabled: true
   phone: 0208 588 4002
-  email: socialvisits.wandsworth@hmps.gsi.gov.uk
+  email: visitsbooking.westmidlands@noms.gsi.gov.uk
   instant_booking: false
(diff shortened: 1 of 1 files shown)


dan-palmer:
Suggest testing this on prod first - perhaps by opening up a console session and sending over an email to APVU, then checking they received it ok. Sometimes GSI will swallow up emails however given that this is a GSI email address I cant see it being an issue.

Something like:
```ruby
class PrisonMailer
  def apvu_test
    mail(to: 'visitsbooking.westmidlands@noms.gsi.gov.uk',
      body: 'Hope you received this',
      content_type: 'text/html',
      subject: 'Testing PVB to APVU')
  end
end

PrisonMailer.apvu_test.deliver_now
```

cc @SteveMarshall

SteveMarshall:
Hey @dan-palmer, I'm not sure I see the need for that: we're already sending emails to APVU using that address, and they're receiving them just fine. Merging, and I'll revert if needs be.

PR #317, opened 2015-09-21
Fix cyclomatic complexity

EGiataganas:


spec/controllers/webhooks_controller_spec.rb
```diff
@@ -122,13 +121,13 @@ def bad_encoding(email)
...
-    before :each do
+    before do
       expect(controller).to receive(:authorized?).and_return(false)
     end
 
-    it "returns a 403 unauthorized status" do
+    specify do
       post :email, authorized(email_from_prison)
-      expect(response.status).to eq(403)
+      expect(response).to have_http_status(:forbidden)
```
SteveMarshall: :+1:

lib/email_validator.rb
```diff
@@ -14,6 +14,7 @@ def validate(record)
 
   private
 
+  # rubocop:disable Metrics/CyclomaticComplexity
```
SteveMarshall: Why're we disabling the check here? Wouldn't it make more sense to follow the pattern you use in `WebhooksController`?

SteveMarshall: As @dan-palmer and I discussed, I'm happy to leave this for #276.

EGiataganas: @SteveMarshall :+1:

PR #318, opened 2015-09-21
Unneeded percent q

tatyree:


dan-palmer:
Might be worth fixing the nasty :fearful:  indentation too - not sure if theres a :cop: in the list that will ask us to address this or not.

PR #324, opened 2015-09-23
Fix line length

EGiataganas:
READY TO MERGED.


SteveMarshall:
It might be nice for some of the earlier commits in this had better title lines than `Fix Line Length`. Can you rebase so they're more like `Fix line length in Heathcheck Controller` and so on?

EGiataganas:
@SteveMarshall will do that once I finish with all. I will go back and add what I fixed.

EGiataganas:
@SteveMarshall changed it!!

.rubocop_todo.yml
```diff
@@ -13,12 +13,41 @@ Lint/HandleExceptions:
...
+    - 'lib/session_guard.rb'
+    - 'lib/slots_manipulator.rb'
+    - 'lib/tasks/browserstack.rake'
+    - 'lib/tasks/maintenance.rake'
+    - 'lib/visitors_manipulator.rb'
+    - 'lib/weekly_confirmations_report.rb'
+    - 'test/mailers/previews/mailer_preview_common.rb'
+    - 'test/mailers/previews/prison_mailer_preview.rb'
+    - 'test/mailers/previews/visitor_mailer_preview.rb'
+    - 'spec/**/*'
```
SteveMarshall: Do we need these line-length exclusions, or are these still to be fixed?

SteveMarshall:
How do we feel about trailing `)`s?

For example:

```ruby
token = encryptor.encrypt_and_sign(
   attach_vo_number(
     remove_unused_slots(
       booked_visit, @confirmation.slot),
     @confirmation))
```

versus:

```ruby
token = encryptor.encrypt_and_sign(
  attach_vo_number(
    remove_unused_slots(
      booked_visit, @confirmation.slot
    ),
    @confirmation
  )
)
```

I prefer the latter, but am open to leaving it as the former if
everyone else is fine with that?

app/controllers/metrics_controller.rb
```diff
@@ -69,7 +79,8 @@ def fortnightly_range
   end
 
   def prison_param
-    (prison = params[:prison]) && Rails.configuration.nomis_ids.include?(prison) && prison
+    (prison = params[:prison]) && prison && \
+      Rails.configuration.nomis_ids.include?(prison)
```
SteveMarshall: I might suggest reworking this to pull out the assignment onto its own line.

app/controllers/webhooks_controller.rb
```diff
@@ -69,7 +72,8 @@ def email_params
   end
 
   def authorized?
-    params[:auth].present? && params[:auth].secure_compare(ENV['WEBHOOK_AUTH_KEY'])
+    params[:auth].present? && params[:auth].
+      secure_compare(ENV['WEBHOOK_AUTH_KEY'])
```
SteveMarshall: Might it make more sense to break this line (and others like it) at `&&`, rather than mid-method-call?

app/helpers/application_helper.rb
```diff
@@ -25,7 +25,8 @@ def display_time_slot(times, glue=' to ')
 
   def display_slot_and_duration(times, glue=', ')
     from, to = times.split('-')
-    [format_time_str(from).strip, (time_from_str(to)-time_from_str(from)).duration].join(glue)
+    [format_time_str(from).strip, (time_from_str(to)-time_from_str(from)).
+      duration].join(glue)
```
SteveMarshall: This might be clearer formatted as:

```ruby
[
  format_time_str(from).strip,
  (time_from_str(to)-time_from_str(from)).duration
].join(glue)
```

app/mailers/visitor_mailer.rb
```diff
@@ -13,7 +13,7 @@ class VisitorMailer < ActionMailer::Base
   attr_reader :visit
   helper_method :visit
 
-  default('List-Unsubscribe' => '<https://www.prisonvisits.service.gov.uk/unsubscribe>')
+  default('List-Unsubscribe' => Rails.configuration.unsubscribe_url)
```
SteveMarshall: :+1:

.rubocop_todo.yml
EGiataganas: @SteveMarshall they are still to be fixed!

app/controllers/metrics_controller.rb
EGiataganas: @SteveMarshall np, I will do a fixup

app/controllers/webhooks_controller.rb
EGiataganas: @SteveMarshall np, I will do a fixup

app/helpers/application_helper.rb
EGiataganas: @SteveMarshall yeah, you're right! Its more cleaner like that, I will do a fixup!

app/mailers/visitor_mailer.rb
EGiataganas: @SteveMarshall I wouldn't have done it without @dan-palmer help and guidance :hand:

app/mailers/visitor_mailer.rb
```diff
@@ -21,27 +21,48 @@ def booking_confirmation_email(visit, confirmation, token)
     @confirmation = confirmation
     @token = token
 
-    mail(from: sender, reply_to: prison_mailbox_email, to: recipient, subject: "Visit confirmed: your visit for #{Date.parse(@slot.date).strftime('%e %B %Y').gsub(/^ /,'')} has been confirmed")
+    mail(from: sender,
+         reply_to: prison_mailbox_email,
+         to: recipient,
+         subject: "Visit confirmed: your visit for \
```
dan-palmer: I believe this will add a load of whitespace to the sentence.

SteveMarshall: I was wondering about that, and meant to ask. There are a whole bunch of these; we should check before merging.

app/models/concerns/person.rb
```diff
@@ -13,7 +13,9 @@ module Person
     validates :date_of_birth,
       presence: true,
       inclusion: {
-        in: ->(_) { MAX_AGE.years.ago.beginning_of_year.to_date..Time.zone.today }
+        in: lambda do |_|
```
dan-palmer: When it comes to lambdas I think we should be consistent and stick with one syntax style.

app/models/prisoner.rb
```diff
@@ -5,12 +5,17 @@ class Prisoner
   attribute :number, String
   attribute :prison_name, String
 
-  validates_format_of :number, with: /\A[a-z]\d{4}[a-z]{2}\z/i, message: "must be a valid prisoner number" # eg a1234aa
-  validates_inclusion_of :prison_name, in: Rails.configuration.prison_data.keys, message: "must be chosen"
+  validates_format_of :number,
+    with: /\A[a-z]\d{4}[a-z]{2}\z/i,
+    message: "must be a valid prisoner number" # eg a1234aa
+  validates_inclusion_of :prison_name,
```
dan-palmer: Not something to handle in this PR worth mentioning.. the preferred rails way for model validations (since RoR 3) is through the use of <a href="http://thelucid.com/2010/01/08/sexy-validation-in-edge-rails-rails-3/">'sexy' validations</a>.

app/models/prisoner.rb
```diff
@@ -5,12 +5,17 @@ class Prisoner
...
+    in: Rails.configuration.prison_data.keys,
+    message: "must be chosen"
   validate :prison_in_service
 
   def prison_in_service
-    if !self.prison_name.blank? && !Rails.configuration.prison_data[self.prison_name]['enabled']
+    if !self.prison_name.blank? && !Rails.configuration.
+        prison_data[self.prison_name]['enabled']
```
dan-palmer: How about something like:

```ruby
if prison_name.present? && 
    !Rails.configuration.prison_data[prison_name]['enabled']
```

app/models/slot.rb
```diff
@@ -11,7 +11,9 @@ class Slot
   validate do
     if index == 0
       errors.add(:date, 'must be given') unless date.present? && date.size == 10
-      errors.add(:times, 'must be given') unless times.present? && times.size == 9
+      unless times.present? && times.size == 9
```
dan-palmer: magic numberzzz :cry:

app/models/visit.rb
```diff
@@ -22,12 +25,14 @@ def validate_amount_of_adults
       errors.add :base, "There must be at least one adult visitor"
     end
     if visitors.count { |v| v.age && v.age >= adult_age } > 3
-      errors.add :base, "You can book a maximum of 3 visitors over the age of #{adult_age} on this visit"
+      errors.add :base, "You can book a maximum of 3 visitors over the age of \
```
dan-palmer: Again.. you'll end up with strings like:

"You can book a maximum of 3 visitors over the age of         18 on this visit"

config/environments/production.rb
```diff
@@ -25,8 +25,13 @@
   config.logstasher.enabled = true
   config.logstasher.suppress_app_log = true
   config.logstasher.log_controller_parameters = true
-  config.logstasher.logger_path = File.join(Rails.root, "log", "logstash_production.json")
-  config.action_mailer.default_url_options = { host: ENV["SERVICE_URL"] || (raise "Missing SERVICE_URL"), protocol: "https" }
+  config.logstasher.logger_path =
+    File.join(Rails.root, "log", "logstash_production.json")
 
+  config.action_mailer.default_url_options = {
+    host: ENV["SERVICE_URL"] || \
```
dan-palmer: Just use 'fetch' - it will raise a `KeyError`

dan-palmer:
```ruby
config.action_mailer.default_url_options = \
    { host: ENV.fetch('SERVICE_URL'), protocol: 'https' }
```

app/mailers/visitor_mailer.rb
dan-palmer: 😂

app/models/visit.rb
dan-palmer: On the other hand it might just get ignored.

app/controllers/application_controller.rb
```diff
@@ -74,9 +75,10 @@ def metrics_logger
   end
 
   def ensure_visit_integrity
-    unless visit && visit.prisoner && visit.prisoner.prison_name.present?
+    unless visit && visit.prisoner && visit.prison_name.present?
       redirect_to edit_prisoner_details_path,
-        notice: 'You need to complete missing information to start or continue your visit request'
+        notice: 'You need to complete missing information to start' \
```
EGiataganas: @dan-palmer @SteveMarshall @threedaymonk @tatyree I just saw that I forgot a space at the end, but tests didn't fail. I believe we want that space right?

threedaymonk: I think this is a prime candidate for pulling out into the locale file, rather than just wrapping the line.

We automated localised notices in peoplefinder; have a look at [application controller](https://github.com/ministryofjustice/peoplefinder/blob/5c8aa8daf3e95bc198fd3f4e22bb3d6ad541f10e/app/controllers/application_controller.rb#L58-L71) and [people controller](https://github.com/ministryofjustice/peoplefinder/blob/5c8aa8daf3e95bc198fd3f4e22bb3d6ad541f10e/app/controllers/people_controller.rb#L62) if you're interested in doing something similar.

threedaymonk: The same goes for the other long strings of English text in this PR, by the way.

EGiataganas: @threedaymonk yeah, I believe it worth to do it like that because some of the line length were similar to that. :disappointed:

app/mailers/visitor_mailer.rb
EGiataganas: @dan-palmer && @SteveMarshall yeah I know that. what do you suggest though?

app/models/prisoner.rb
EGiataganas: @dan-palmer :+1:  done

app/models/visit.rb
EGiataganas: @dan-palmer the spaces now will be ok?

app/models/visit.rb
```diff
@@ -22,12 +25,14 @@ def validate_amount_of_adults
       errors.add :base, "There must be at least one adult visitor"
     end
     if visitors.count { |v| v.age && v.age >= adult_age } > 3
-      errors.add :base, "You can book a maximum of 3 visitors over the age of #{adult_age} on this visit"
+      errors.add :base, "You can book a maximum of 3 visitors over the age of"\
```
EGiataganas: @dan-palmer the spaces now will be ok?

dan-palmer: As @threedaymonk suggested its worth considering pulling these out into the locale file.

SteveMarshall: Yeah, let's rip 'em all out.

app/mailers/visitor_mailer.rb
threedaymonk: Does the lack of surrounding `<` `>` matter?

app/mailers/visitor_mailer.rb
```diff
@@ -21,27 +21,49 @@ def booking_confirmation_email(visit, confirmation, token)
...
     @token = token
 
-    mail(from: sender, reply_to: prison_mailbox_email, to: recipient, subject: "Visit confirmed: your visit for #{Date.parse(@slot.date).strftime('%e %B %Y').gsub(/^ /,'')} has been confirmed")
+    mail(
+      from: sender,
+      reply_to: prison_mailbox_email,
+      to: recipient,
+      subject: "Visit confirmed: your visit for " \
+       "#{Date.parse(@slot.date).strftime('%e %B %Y').gsub(/^ /,'')} "\
+       "has been confirmed"
```
threedaymonk: The subject could be localised, too.

dan-palmer: Further to this I <i>think</i> you can just remove having to put the subject and the localisation path in the mail. ActionMailer will do the lookup for you.

dan-palmer: (@EGiataganas)
```ruby
mail(from: sender, reply_to: prison_mailbox_email, to: recipient)
```

threedaymonk: It can localise the subject automatically: it's *name of mailer* → *name of method* → *subject*, *viz.*:

```yaml
en:
  visitor_mailer:
    booking_confirmation_email:
      subject: Whatever
```

However, I'm not sure how interpolated values are passed (or if they can be).

dan-palmer: Friday brain didn't think of that.. good point!

dan-palmer: <a href="http://guides.rubyonrails.org/i18n.html#translations-for-action-mailer-e-mail-subjects">default_i18n_subject</a> :zap:

app/mailers/visitor_mailer.rb
dan-palmer: The config value includes the `<` `>`.

They do appear to be necessary - well according to the examples anyway. (http://www.faqs.org/rfcs/rfc2369.html)

app/controllers/geckoboard_controller.rb
```diff
@@ -24,4 +24,11 @@ def confirmed_bookings
             ]
     }
   end
+
+  private
+
+  def leaderboard_report
+    LeaderboardReport.new(0.95, ApplicationHelper.
+                                  instance_method(:prison_name_for_id))
```
threedaymonk: My personal preference for things like this would be:

```ruby
LeaderboardReport.new(
  0.95,
  ApplicationHelper.instance_method(:prison_name_for_id)
)
```

(I don't really like lining up indentation as it results in volatile changesets.)

threedaymonk: Separately, what is 0.95? Can we make it a constant?

threedaymonk:
@SteveMarshall I feel the same as you about trailing `)`s: I think it's more consistent with the rest of the codebase to put them on a new line.

app/controllers/geckoboard_controller.rb
SteveMarshall: :+1: on all of @threedaymonk's suggestions here.

app/controllers/metrics_controller.rb
```diff
@@ -7,13 +7,16 @@ def index
...
+        VisitMetricsEntry.deferred, 3.days, fortnightly_range)
     end
 
     respond_to do |format|
       format.html
       format.csv do
-        render text: CSVFormatter.new(@nomis_ids, ApplicationHelper.instance_method(:prison_name_for_id)).generate(@dataset)
+        render text: CSVFormatter.new(
+          @nomis_ids, ApplicationHelper.instance_method(:prison_name_for_id)).
+        generate(@dataset)
```
threedaymonk: I think this is a good indication of where putting `)` on the next line helps. At a first read, it's not clear that `generate` follows on, whereas

```ruby
).generate(@dataset)
```

is more obvious.

app/controllers/prisoner_details_controller.rb
```diff
@@ -23,7 +30,8 @@ def update
   def prisoner_params
     dob = [:'date_of_birth(3i)', :'date_of_birth(2i)', :'date_of_birth(1i)']
     if params[:date_of_birth_native].present?
-      params[:prisoner][:date_of_birth] = Date.parse(params[:date_of_birth_native])
+      params[:prisoner][:date_of_birth] = Date.parse(
+        params[:date_of_birth_native])
```
threedaymonk: I think this kind of thing works better as

```ruby
params[:prisoner][:date_of_birth] =
  Date.parse(params[:date_of_birth_native])
```

because it keeps the parameters with the method.

app/controllers/webhooks_controller.rb
```diff
@@ -33,7 +33,7 @@ def handle_unknown_recipient!
 
   def handle_blacklisted!
     if BLACKLISTED_SENDERS.include? parsed_email.from.local.downcase
-      logger.error "Blacklisted sender detected. Skipping message.\n #{parsed_email.inspect}"
+      logger.error :blacklisted
```
threedaymonk: Do these actually get localised? What about the interpolated values?

(I'm not sure whether dumping the email into the logs is actually a good idea, but that's another question!)

app/controllers/webhooks_controller.rb
```diff
@@ -69,7 +69,8 @@ def email_params
   end
 
   def authorized?
-    params[:auth].present? && params[:auth].secure_compare(ENV['WEBHOOK_AUTH_KEY'])
+    params[:auth].present? && \
```
threedaymonk: You don't need `\` here, because the parser knows to continue after `&&`.

app/helpers/application_helper.rb
```diff
@@ -53,13 +56,15 @@ def conditional_text(variable, prefix = '', suffix = '')
   end
 
   def prison_name_for_id(nomis_id)
-    @nomis_id_to_prison ||= Rails.configuration.prison_data.inject({}) do |h, (prison_name, prison_data)|
-      if prison_data['enabled']
-        h.update(prison_data['nomis_id'] => prison_name)
-      else
-        h
+    @nomis_id_to_prison ||= \
```
threedaymonk: `\` is unnecessary here.

app/helpers/visit_helper.rb
```diff
@@ -55,12 +53,13 @@ def adult_age
 
   def prison_url(visit)
     data = prison_data(visit)
-    slug = data['finder_slug'] ? data['finder_slug'] : visit.prisoner.prison_name.parameterize
+    slug = data['finder_slug'] ? data['finder_slug'] : visit.prisoner.
+      prison_name.parameterize
```
threedaymonk: I think we could use `fetch` instead and avoid some repetition:

```ruby
slug = data.fetch('finder_slug') { visit.prisoner.prison_name.parameterize }
```

app/helpers/visit_helper.rb
```diff
@@ -89,7 +88,8 @@ def when_to_expect_reply
   end
 
   def prison_specific_id_requirements(prison_name)
-    nomis_id = Rails.configuration.prison_data.fetch(prison_name).fetch(:nomis_id)
+    nomis_id = \
```
threedaymonk: `\` is not needed.

app/mailers/prison_mailer.rb
```diff
@@ -28,7 +28,14 @@ def booking_request_email(visit, token)
...
 
-    mail(from: sender, reply_to: user, to: recipient, subject: "Visit request for #{@visit.prisoner.full_name} on #{Date.parse(@visit.slots.first.date).strftime('%A %e %B')}")
+    mail(from: sender,
+         reply_to: user,
+         to: recipient,
+         subject: default_i18n_subject(
+           full_name: @visit.prisoner.full_name,
+           date: Date.parse(@visit.slots.first.date).strftime('%A %e %B')
+         )
+        )
```
threedaymonk: Could we also indent the `from:` so it's all a consistent two spaces?

app/models/prisoner.rb
```diff
@@ -5,12 +5,21 @@ class Prisoner
...
+    in: Rails.configuration.prison_data.keys,
+    message: "must be chosen"
+  }
+
   validate :prison_in_service
 
   def prison_in_service
-    if !self.prison_name.blank? && !Rails.configuration.prison_data[self.prison_name]['enabled']
+    if !self.prison_name.blank? &&
+      !Rails.configuration.prison_data[self.prison_name]['enabled']
```
threedaymonk: This could be shorter without the redundant `self`:

```ruby
if prison_name.present? && !Rails.configuration.prison_data[prison_name]['enabled']
```

app/mailers/prison_mailer.rb
EGiataganas: I tried it, but then rubocop will say "Space inside parentheses detected." so what I can suggest is 

```ruby
mail(
      from: sender,
      reply_to: user,
      to: recipient,
      subject: default_i18n_subject(
        full_name: @visit.prisoner.full_name,
        date: Date.parse(@visit.slots.first.date).strftime('%A %e %B')
      )
    )
```

app/models/prisoner.rb
EGiataganas: @threedaymonk :+1:

app/models/visit.rb
```diff
@@ -12,7 +12,10 @@ class Visit
   attribute :vo_number, String
 
   validates_presence_of :visit_id
-  validates_size_of :slots, within: 1..MAX_SLOTS, on: :date_and_time, message: 'must be at least one and at most three'
+  validates_size_of :slots,
+    within: 1..MAX_SLOTS,
+    on: :date_and_time,
+    message: 'must be at least one and at most three'
```
threedaymonk: I have three problems with the existing code, which would be nice to fix while we're here.

1. It's strangely named (`size_of` is an alias for `length_of`, and `within` for `in`).
2. It uses the old style `validates_bar :foo` instead of `validate :foo, bar: …`.
3. I can't work out what `on: :date_and_time` does! Does it do anything?

```ruby
validate :slots, length: { in: 1..MAX_SLOTS }
```

Is probably enough, with the message moved to `en.yml`. We then get too short and too long for free, instead of this vague message:

```yaml
en:
  activemodel:
    errors:
      models:
        visit:
          attributes:
            slots:
              too_short: 'must be at least one'
              too_long: 'must be at most three'
```

app/models/visit.rb
threedaymonk: I think these errors would be more appropriate on `:visitors` instead of `:base`. To localise them, it's something like this:

```ruby
errors.add :visitors, :max_3_adults, adult_age: adult_age
```

And then the appropriate localisation key will be something like:

```yaml
en:
  activemodel:
    errors:
      models:
        visit:
          attributes:
            visitors:
              max_3_adults: 'You can book a maximum of 3 visitors over the age of %{adult_age} on this visit'
```

config/application.rb
```diff
@@ -23,7 +25,9 @@ class Application < Rails::Application
       back-office.css
       *.png
     )
-    config.prison_data = YAML.load_file(ENV['PRISON_DATA_FILE'] || File.join(Rails.root, 'config', 'prison_data_staging.yml')).with_indifferent_access
+    config.prison_data = YAML.load_file(ENV['PRISON_DATA_FILE'] || \
```
threedaymonk: We don't need `\` after `||` (indeed, after any binary operator).

config/initializers/markdown_handler.rb
```diff
@@ -5,10 +5,13 @@ def self.call(template)
...
     SOURCE
   end
 end
 
-ActionView::Template.register_template_handler(:md, MarkdownTemplateHandler)
-ActionView::Template.register_template_handler(:markdown, MarkdownTemplateHandler)
+ActionView::Template.register_template_handler(
+  :md, MarkdownTemplateHandler)
+ActionView::Template.register_template_handler(
+  :markdown, MarkdownTemplateHandler)
```
threedaymonk: I think I'd prefer this, if it fits:

```ruby
ActionView::Template.
  register_template_handler(:markdown, MarkdownTemplateHandler)
```

lib/calculated_metrics.rb
```diff
@@ -47,34 +60,48 @@ def percent_rejected(prison, reason=nil)
...
+                      OVER (PARTITION BY nomis_id ORDER BY ?
+                    ) AS percentile
                      FROM visit_metrics_entries WHERE ? IS NOT NULL
                      AND requested_at > ?::date
                      AND processed_at <= ?::date),
-     top_percentiles AS (SELECT nomis_id, ?, rank() OVER (PARTITION BY nomis_id ORDER BY ?)
+     top_percentiles AS (SELECT nomis_id, ?, rank()
+                          OVER (PARTITION BY nomis_id ORDER BY ?)
                      FROM percentiles WHERE percentile >= ?)
 SELECT nomis_id, ? FROM top_percentiles WHERE rank = 1
```
threedaymonk: This SQL could be indented just like the rest of the code: it doesn't have to start at the first column.

Also, I think it's nice to use `<<-SQL` … `SQL` as a delimiter: some editors can then highlight the inner language properly:

```ruby
query = <<-SQL
  WITH percentiles AS (
    SELECT nomis_id, ?, cume_dist()
      OVER (PARTITION BY nomis_id ORDER BY ?
    ) AS percentile
  …
SQL

@model.find_by_sql [
  query, column, column, …
]
```

lib/weekly_confirmations_report.rb
```diff
@@ -61,9 +60,11 @@ def refresh
 
   def csv
     CSV.generate(headers: true) do |csv|
-      csv << ['Prison'] + week_range.map { |weekno| @start_of_year + weekno * 7 } + ['NOMIS ID']
+      csv << ['Prison'] + week_range.
+        map { |weekno| @start_of_year + weekno * 7 } + ['NOMIS ID']
       @dataset.keys.sort.each do |prison|
-        csv << [@prison_labeling_function.call(prison)] + week_range.map { |weekno| @dataset[prison][weekno] } + [prison]
+        csv << [@prison_labeling_function.call(prison)] + week_range.
+          map { |weekno| @dataset[prison][weekno] } + [prison]
```
threedaymonk: I think it's better to split on `+` than `.` here: it breaks up higher-level components of the line.

app/models/visit.rb
```diff
@@ -12,7 +12,10 @@ class Visit
   attribute :vo_number, String
 
   validates_presence_of :visit_id
```
EGiataganas: I will also change it to 

```ruby
validates :visit_id, presence: true
```

app/models/visit.rb
EGiataganas: @threedaymonk :+1: The `on: :date_and_time` I believe it is necessary, because without it we are having some failing test:

```ruby
rspec './spec/features/visitor_information_page_spec.rb[1:2:5:1]' # visitor enters visitor information and they fill out all fields for 4 additional visitors displays the calendar
rspec './spec/features/visitor_information_page_spec.rb[1:2:4:1]' # visitor enters visitor information and they fill out all fields for 3 additional visitors displays the calendar
rspec './spec/features/visitor_information_page_spec.rb[1:2:2:1]' # visitor enters visitor information and they fill out all fields for 1 additional visitors displays the calendar
rspec ./spec/features/visitor_information_page_spec.rb:25 # visitor enters visitor information and they fill out all fields for one visitor displays the calendar
rspec './spec/features/visitor_information_page_spec.rb[1:2:3:1]' # visitor enters visitor information and they fill out all fields for 2 additional visitors displays the calendar
rspec './spec/features/visitor_information_page_spec.rb[1:2:6:1]' # visitor enters visitor information and they fill out all fields for 5 additional visitors displays the calendar
rspec ./spec/features/unexpected_journey_spec.rb:72 # unexpected journeys through the application session expires on the confirmation page
rspec ./spec/features/unexpected_journey_spec.rb:56 # unexpected journeys through the application session expires while entering visit details
rspec ./spec/controllers/visitors_details_controller_spec.rb:52 # VisitorsDetailsController given valid visitor information updates visitor information
rspec './spec/controllers/visitors_details_controller_spec.rb[1:4:4:1:1]' # VisitorsDetailsController behaves like a visitor data manipulator with invalid data given only a single child visitor when the prison assumes each adult to be 18 or older allows the visitor to proceed
rspec './spec/controllers/visitors_details_controller_spec.rb[1:4:4:2:1]' # VisitorsDetailsController behaves like a visitor data manipulator with invalid data given only a single child visitor when the prison assumes each adult to be some other age allows the visitor to proceed
rspec ./spec/models/visit_spec.rb:41 # Visit is valid with three or fewer adult visitors
rspec ./spec/models/visit_spec.rb:16 # Visit restricts the number of visitors
rspec ./spec/models/visit_spec.rb:56 # Visit given a prison which treats a child as an adult for seating purposes requires at least one real adult
rspec ./spec/features/datepicker_spec.rb:36 # visitor selects a date Choosing a date more than 14 days from now
rspec ./spec/features/datepicker_spec.rb:41 # visitor selects a date Booking a valid slot
rspec ./spec/features/datepicker_spec.rb:30 # visitor selects a date Choosing a date more than 3 working days from now
rspec ./spec/features/datepicker_spec.rb:24 # visitor selects a date Choosing a date in the past
```

Also I found that is using it [slots_manipulator_line 18](https://github.com/ministryofjustice/prison-visits/blob/master/lib/slots_manipulator.rb#L18), [visit_spec_line 30](https://github.com/ministryofjustice/prison-visits/blob/master/spec/models/visit_spec.rb#L30),  [visit_spec_line 34](https://github.com/ministryofjustice/prison-visits/blob/master/spec/models/visit_spec.rb#L34) and  [visit_spec_line 38](https://github.com/ministryofjustice/prison-visits/blob/master/spec/models/visit_spec.rb#L38) if I understand correct.

By refactoring it to 
```ruby
validates :slots, length: {
    in: 1..MAX_SLOTS,
    on: :date_and_time
  }
```
all the test are passing.

app/controllers/prisoner_details_controller.rb
EGiataganas: @threedaymonk :+1:

config/initializers/markdown_handler.rb
EGiataganas: @threedaymonk :+1:

app/mailers/prison_mailer.rb
threedaymonk: That's exactly what I meant!

EGiataganas: @threedaymonk sorry then, my bad... :stuck_out_tongue_winking_eye:

app/controllers/confirmations_controller.rb
```diff
@@ -24,21 +27,39 @@ def new
...
     end
 
     if @confirmation.slot_selected?
-      token = encryptor.encrypt_and_sign(attach_vo_number(remove_unused_slots(booked_visit, @confirmation.slot), @confirmation))
-      VisitorMailer.booking_confirmation_email(booked_visit, @confirmation, token).deliver_later
+      token = encryptor.encrypt_and_sign(
+        attach_vo_number(
+          remove_unused_slots(
+            booked_visit, @confirmation.slot),
+          @confirmation))
```
threedaymonk: Can we put `)` on its own line everywhere?

app/controllers/application_controller.rb
```diff
@@ -20,7 +20,8 @@ def reject!
   end
 
   def reject_untrusted_ips!
-    unless Rails.configuration.permitted_ips_for_confirmations.include?(request.remote_ip)
+    unless Rails.configuration.permitted_ips_for_confirmations.
+                  include?(request.remote_ip)
```
threedaymonk: This indentation seems a bit excessive (and a bit odd).

app/controllers/metrics_controller.rb
```diff
@@ -7,13 +7,17 @@ def index
       @dataset = CalculatedMetrics.new(VisitMetricsEntry.deferred, 3.days)
     else
       @start_date, @end_date = fortnightly_range.first, fortnightly_range.last
-      @dataset = CalculatedMetrics.new(VisitMetricsEntry.deferred, 3.days, fortnightly_range)
+      @dataset = CalculatedMetrics.new(
+        VisitMetricsEntry.deferred, 3.days, fortnightly_range)
```
threedaymonk: `)`

app/controllers/feedbacks_controller.rb
```diff
@@ -23,6 +23,7 @@ def create
   private
 
   def feedback_params
-    params.require(:feedback).permit(:referrer, :text, :email, :user_agent, :prison)
+    params.require(:feedback).permit(
+      :referrer, :text, :email, :user_agent, :prison)
```
threedaymonk: Either putting the `)` on a newline or putting the whole `.permit(…)` on its own line would be better, I think.

app/helpers/application_helper.rb
```diff
@@ -53,13 +56,15 @@ def conditional_text(variable, prefix = '', suffix = '')
...
-      else
-        h
+    @nomis_id_to_prison ||=
+      Rails.configuration.prison_data.
+        inject({}) do |h, (prison_name, prison_data)|
+        if prison_data['enabled']
+          h.update(prison_data['nomis_id'] => prison_name)
+        else
+          h
+        end
```
threedaymonk: The indentation here is confusing. But so's the code! I think it might be clearer as:

```ruby
@nomis_id_to_prison ||= Hash[
  Rails.configuration.prison_data.
    select { |_, data| data['enabled'] }.
    map { |name, data| [data['nomis_id'], name] }
]
```

If we were on an up-to-date version of Ruby, we could use `.to_h` instead of `Hash[]`. But we're not. :cry:

app/mailers/visitor_mailer.rb
```diff
@@ -21,21 +21,40 @@ def booking_confirmation_email(visit, confirmation, token)
...
+      )
+    )
   end
 
   def booking_rejection_email(visit, confirmation)
     @visit = visit
     @confirmation = confirmation
 
-    mail(from: sender, reply_to: prison_mailbox_email, to: recipient, subject: "Visit cannot take place: your visit for #{first_date.strftime('%e %B %Y').gsub(/^ /,'')} could not be booked")
+    mail(from: sender,
```
threedaymonk: Can we indent this two spaces too?

app/mailers/visitor_mailer.rb
```diff
@@ -21,21 +21,40 @@ def booking_confirmation_email(visit, confirmation, token)
...
+         )
+        )
   end
 
   def booking_receipt_email(visit, token)
     @visit = visit
     @token = token
 
-    mail(from: sender, reply_to: prison_mailbox_email, to: recipient, subject: "Not booked yet: we've received your visit request for #{first_date.strftime('%e %B %Y').gsub(/^ /,'')}")
+    mail(from: sender,
```
threedaymonk: And this.

app/models/prisoner.rb
```diff
@@ -5,12 +5,21 @@ class Prisoner
   attribute :number, String
   attribute :prison_name, String
 
-  validates_format_of :number, with: /\A[a-z]\d{4}[a-z]{2}\z/i, message: "must be a valid prisoner number" # eg a1234aa
-  validates_inclusion_of :prison_name, in: Rails.configuration.prison_data.keys, message: "must be chosen"
+  validates :number, format: {
+    with: /\A[a-z]\d{4}[a-z]{2}\z/i,
+    message: "must be a valid prisoner number" # eg a1234aa
```
threedaymonk: You can move these messages to `en.yml`.

config/environments/production.rb
```diff
@@ -25,8 +25,13 @@
   config.logstasher.enabled = true
   config.logstasher.suppress_app_log = true
   config.logstasher.log_controller_parameters = true
-  config.logstasher.logger_path = File.join(Rails.root, "log", "logstash_production.json")
-  config.action_mailer.default_url_options = { host: ENV["SERVICE_URL"] || (raise "Missing SERVICE_URL"), protocol: "https" }
+
+  config.logstasher.logger_path =
+    File.join(Rails.root, "log", "logstash_production.json")
+
+  config.action_mailer.default_url_options = \
```
threedaymonk: Unnecessary `\`

lib/detailed_windowed_metrics.rb
```diff
@@ -2,7 +2,8 @@ class DetailedWindowedMetrics < DetailedMetrics
   def initialize(model, nomis_id, date_range)
     super(model, nomis_id)
     @scoped_model_for_histograms = @scoped_model.after(date_range.last)
-    @scoped_model = @scoped_model.after(date_range.first).before(date_range.last)
+    @scoped_model = \
```
threedaymonk: `\` is not needed

lib/tasks/maintenance.rake
```diff
@@ -19,6 +22,7 @@ namespace :maintenance do
   end
 
   task :update_bank_holidays do
-    `curl -k https://www.gov.uk/bank-holidays.json -o config/bank-holidays.json`
+    `curl -k https://www.gov.uk/bank-holidays.json -o \
+    config/bank-holidays.json`
```
threedaymonk: We could replace this entirely with [sh](http://rake.rubyforge.org/FileUtils.html#method-i-sh).

app/controllers/application_controller.rb
EGiataganas: @threedaymonk sorry :cry:

lib/visitors_manipulator.rb
```diff
@@ -24,17 +24,22 @@ def update
...
 
+  def max_visitors
+    if visit.visitors.size < Visit::MAX_VISITORS
+      visit.visitors << model_class.new
+      redirect_to this_path
+    else
+      flash[:notice] = I18n.t(:max_visitors, scope: 'controllers.shared')
+      redirect_to this_path
+    end
+  end
```
threedaymonk: I love that you've extracted this, but I think the name is misleading. It seems to do two things:

1. ensure that there is a blank visitor to add a new one, if you haven't reached the maximum, or
2. flash a notice if you have reached the maximum, then
3. redirect.

But as it always redirects identically, perhaps that doesn't belong in the method. Then this method would be something like:

```ruby
def prepare_next_visitor
  if visit.visitors.size < Visit::MAX_VISITORS
    visit.visitors << model_class.new
  else
    flash[:notice] = I18n.t(:max_visitors, scope: 'controllers.shared')
  end
end
```

and we'd call redirect after returning.

lib/visitors_manipulator.rb
```diff
@@ -62,6 +67,7 @@ def visitors_params
           "#{date_of_birth.inspect}\nexception: #{e}"
       end
     end
-    ParamUtils.trim_whitespace_from_values(params.require(:visit).require(:visitor))
+    ParamUtils.trim_whitespace_from_values(
+      params.require(:visit).require(:visitor))
```
threedaymonk: `)`

app/models/prisoner.rb
```diff
@@ -5,12 +5,21 @@ class Prisoner
...
+  }
+
+  validates :prison_name, inclusion: {
+    in: Rails.configuration.prison_data.keys,
+    message: "must be chosen"
```
EGiataganas: @threedaymonk I will also add that

lib/visitors_manipulator.rb
EGiataganas: @threedaymonk thank you once again! :+1: :smile:

lib/visitors_manipulator.rb
EGiataganas: @threedaymonk yeah, you re right and by moving redirect after the `if` then we won't have duplications.

EGiataganas:
Tested on preprod, but email processing is not working, so I can't test the staff admin side. Blocked until queue processing is fixed.

SteveMarshall:
Emails should be working on preprod again now, @EGiataganas. Test when ready! (And rebase against master, perhaps?)

EGiataganas:
@SteveMarshall :+1: Thank you very much. Will do that now.

config/locales/en.yml
```diff
@@ -20,6 +27,14 @@ en:
               too_short: 'must include area code'
       messages:
         four_digit_year: 'year must be four digits'
```
EGiataganas: @threedaymonk is it correct like that? or should be something like
```ruby
models:
 concerns:
    person:
      messages:
         four_digit_year: 'year must be four digits'
```
I run the test like that and they passed. Then I delete those lines:
```ruby
       messages:
         four_digit_year: 'year must be four digits'
```
And run the tests again and they pass again. So I am not sure if we have to change it, as we added more details to locales.

EGiataganas:
@SteveMarshall @threedaymonk Tested on preprod & email processing is working.

PR #325, opened 2015-09-23
Remove documentary comments/associated whitespace

tatyree:
I started to do this as part of the rubocop LineLength TODO. On
reflection, it seemed cleaner to do it on a separate branch as not all
of them were excessively long.  Nevertheless, we don't want them in the
codebase.  Where removal would result in an empty file, as with
config/initializers/backtrace_silencers.rb, I have left the opening
comment (almost always a warning to restart the server if the file
changes).


config/initializers/inflections.rb
```diff
@@ -1,16 +1 @@
 # Be sure to restart your server when you modify this file.
```
SteveMarshall: What about this one?

config/initializers/backtrace_silencers.rb
```diff
@@ -1,7 +1 @@
 # Be sure to restart your server when you modify this file.
```
SteveMarshall: What about this?

SteveMarshall:
Other than those two comments, this looks great. Remove all the things!

tatyree:
@SteveMarshall from my opening comment: "Where removal would result in an empty file, as with
config/initializers/backtrace_silencers.rb, I have left the opening
comment..." :)

SteveMarshall:
@tatyree D'oh! Is there any reason not to remove the files? Would it break anything?

tatyree:
@SteveMarshall I'll try and see what happens.

tatyree:
@SteveMarshall Looks GTG. Specs pass without them, removed and rebased into original commit.

threedaymonk:
caedete eos!

PR #327, opened 2015-09-24
Fix symbolproc

tatyree:
As per the rubocop docs: 

http://www.rubydoc.info/gems/rubocop/0.26.1/RuboCop/Cop/Style/SymbolProc

Not too many of these, thankfully.


dan-palmer:
@SteveMarshall regarding the `browserstack` rake task - do we even need this? I took a look at the jenkins job & can see that its never really worked, appears to just be failing silently at the moment.

tatyree:
@dan-palmer @SteveMarshall I'm happy to take it out, but shouldn't we do it as part of a separate issue/task?

lib/tasks/browserstack.rake
```diff
@@ -41,11 +41,11 @@ namespace :browserstack do
         # We're in a subprocess here - set the environment variable BS_BROWSER to the desired browser configuration.
         ENV['BS_BROWSER'] = browser.to_json
 
-        test_label = ['os', 'os_version', 'browser', 'browser_version'].map { |k| browser[k] }.join('_')
+        test_label = ['os', 'os_version', 'browser', 'browser_version'].map(&:browser).join('_')
 
         system("rspec spec/features --format RspecJunitFormatter --out '#{test_label}.xml'")
       end
-      exitstatus = results.count { |e| !e }
+      exitstatus = results.count(&:'!e')
```
dan-palmer: Surely this doesn't work - you'll just end up with a `NoMethodError`.

Anyways, the intention of this line seems to be to count the failures & if any exist cause the program to exit with failure. If we assume the 'bad' results are falsey values couldn't we just do:

```ruby
results.each { |result| abort unless result }
```
& do away with the ensure as well 😕

SteveMarshall:
@dan-palmer It'd be nice to get it working, I guess… But I'm also happy to just kill it (but, yeah, separate PR).

tatyree:
Ready to be merged.

lib/tasks/browserstack.rake
dan-palmer: @tatyree *bump*

tatyree: @dan-palmer I'll take a look.  Is it worth bothering with at all if it's never worked and we plan to take it out soon anyway?

dan-palmer: I think it should still be valid code until its removed. We might end up keeping it and just getting it to work correctly in which case it'll cause somebody some pain.

tatyree: @dan-palmer agreed.  The whole thing is making my brain hurt. 

So, you're saying the actual error count does't matter so long as we fail fast when we encounter the first error?

PR #328, opened 2015-09-24
Fix StructInheritence violations 

tatyree:
A good explaination here:

http://thepugautomatic.com/2013/08/struct-inheritance-is-overused/


tatyree:
Changing to an initializer bolierplate worked fine for prison_day.rb but doesn't seem to work for prison_schedule.  It's the most straightforward resolution, but involved the most extra code (as compared to something like the `attr_extras` gem mentioned in the cited blog post.  Just wanted to give everyone a chance to comment before I went ahead and worked out why prison_schedule.rb doesn't like the initializer (realising that this carries the assumption that `attr_exrtras` would work...).

dan-palmer:
Totes justified in making these changes :+1: 

Regarding the `prison_schedule` though I don't see why you'd have a problem doing:

```ruby
class PrisonSchedule
  attr_reader :prison

  def initialize(prison)
    @prison = prison
  end
  # blah blah blah
end
```

tatyree:
@dan-palmer Because I was being an idiot, that's why :boom:

Fixed and ready for review.

dan-palmer:
:+1:

.rubocop_todo.yml
```diff
@@ -292,6 +284,13 @@ Style/IndentHash:
   Enabled: false
 
 # Offense count: 1
+# Cop supports --auto-correct.
+# Configuration parameters: Width.
+Style/IndentationWidth:
+  Exclude:
+    - 'lib/prison_schedule.rb'
```
SteveMarshall: Where's this come from? Looks like you've got [a 1-space indent](#diff-6bff082cf145ad1fbaeecb4fd4df90f5R5):

```ruby
  def initialize(prison)
   @prison = prison
  end
```

tatyree:
bump for merge once circle it finished...

.rubocop_todo.yml
SteveMarshall: You still need to fix the one-space indent in `prison_schedule.rb`, I think?

tatyree: Oops.  Missed that.  Sec.

tatyree:
@SteveMarshall that one-spacer is fixed.

SteveMarshall:
Friday afternoon :ship: time!

PR #330, opened 2015-09-25
Remove instant booking

threedaymonk:
This isn't used, it adds a load of complexity, and if we do introduce instant booking in the future it won't be in this form.


spec/support/killswitch_support.rb
```diff
@@ -1,11 +1,4 @@
 RSpec.shared_examples "a killswitch-enabled controller" do
```
dan-palmer: Could this whole file just be shown the :door: ?

threedaymonk: Almost certainly! I intend to do so.

SteveMarshall:
:fire::fire::fire::fire::fire::fire::fire:

app/views/deferred/confirmations/show.html.haml
```diff
@@ -12,4 +12,4 @@
         %p Please notify us now if your social visits slots will change due to Easter Bank Holidays. Please email changes to prison-visits-feedback@digital.justice.gov.uk.
 
         %p
-          %a(href="/staff/changes") Read about recent updates
+          = link_to 'Read about recent updates', changes_path
```
dan-palmer: :+:

threedaymonk:
Tested on preprod.

app/models/visit_metrics_entry.rb
```diff
@@ -1,6 +1,6 @@
 class VisitMetricsEntry < ActiveRecord::Base
   validates_presence_of :visit_id, :nomis_id, :requested_at
-  validates_inclusion_of :kind, in: %w{deferred instant}
+  validates_inclusion_of :kind, in: %w{deferred}
```
SteveMarshall: Can we get rid of this? I guess that'd involve a migration?

threedaymonk: It would also involve some changes to metrics, which don't have any test coverage, and which I'd rather leave entirely alone until we replace them with something working.

SteveMarshall: :+1:

spec/models/visit_metrics_entry_spec.rb
```diff
@@ -2,6 +2,8 @@
 
 RSpec.describe VisitMetricsEntry do
   it "requires mandatory fields" do
+    # TODO: Rewrite this spec or throw it away.
+
```
SteveMarshall: I think, given this is in a metrics-related file, and we're going to be looking at (read: trashing) the metrics soon, this comment is probably unnecessary?

SteveMarshall:
LGTM! :fire:

spec/models/visit_metrics_entry_spec.rb
threedaymonk: Fair point

EGiataganas:
:+1:

PR #333, opened 2015-09-25
Log visitor DOB parse error rescues

tatyree:
As per the .rubocop_todo.yml.


lib/visitors_manipulator.rb
```diff
@@ -57,8 +57,8 @@ def visitors_params
       end
       begin
         visitor[:date_of_birth] = Date.new(*date_of_birth.reverse)
-      rescue ArgumentError
-        # NOOP
+      rescue ArgumentError => e
+        logger.error "Issue with visitor DOB detected ( VisitorsManipulator): #{date_of_birth.inspect}\nexception: #{e}"
```
SteveMarshall: Might be nice if this line were broken up a bit, and didn't have the space in `( VisitorsManipulator)`.

PR #336, opened 2015-09-28
Allow deserialisation of legacy marshalled data

threedaymonk:
Instead of a database, this application passes objects around as marshalled data. The renaming of `Deferred::Visitor` to `Visitor` meant that previously-serialised `Deferred::Visitor` instances could not be deserialised. This affects live sessions (although these die after 20 minutes), links to return to the booking, and links for admin staff.

Adding back the namespaced class means that both class names will continue to work. Only `Visitor` instances will be created in future, and we can remove this namespaced class in due course.


threedaymonk:
Still to check: does anything use implicit routes? I think probably not, because there's no RESTful or resourceful routing in this application.

threedaymonk:
In answer to my query: yes, but only in three places, and nowhere for a `Visitor`, which was the only namespaced model.

SteveMarshall:
As we discussed, this should use a URL serialiser that isn't tied to our internal objects. That way, we can support the old object deserialisation until we see no more evidence of their use in the logs (and then remove it), and have a serialisation format that doesn't lock us in to the new objects.

We also discussed whether we should version the new serialisation format and concluded that, no, versions of serialisation schemas would become just another thing to manage and, given that the core data in the system is unlikely to change in the near future, that's overhead we probably don't need.

spec/features/deserialization_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+        VisitMetricsEntry.create! hash
+      end
+      allow_any_instance_of(ApplicationController).
+        to receive(:reject_untrusted_ips!)
+    end
+
+    scenario 'booking receipt' do
+      link = data.fetch('booking_receipt_path')
+      visit link
+      expect(status).to eq(200)
```
SteveMarshall: Might be nice if these used [the `have_http_status` matcher](https://relishapp.com/rspec/rspec-rails/docs/matchers/have-http-status-matcher).

config/routes.rb
```diff
@@ -67,6 +67,7 @@
 
   # Legacy URLs
   get "/prisoner-details", to: redirect("/prisoner")
+  get "/deferred/confirmation/new" => "confirmations#new"
```
threedaymonk: Someone processing booking requests follows a link from an email and ends up on the URL they expected from the link.

Whilst Rails's routing does support redirection, it discards the query-string parameters, so we'd have to do something in a controller. I'm not sure it's worth that additional complexity, and I'm not convinced it's any improvement from a user's point of view.

SteveMarshall: Yeah, I figured it might be something like that. Leave it as-is :+1:

SteveMarshall:
The other thing that's missing here is some way for us to track usage of `Deferred::Visitor`. Probably we should just be able to log things out and graph the number of times it's logged in Kibana (but I sure as hell can't see an explicit way to log directly the information we want).

spec/features/deserialization_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  let(:data) {
+    YAML.load(
+      File.read(
+        File.expand_path('../../fixtures/deserialization_data.yml', __FILE__)
+      )
+    )
+  }
+
```
tatyree: Minor style thing--the ruby-style guide prefers do...end for multiline blocks.  I'm not sure whether or no rubocop will alert on this...

threedaymonk: I know this passes: I run Rubocop every time I run rake :-)

However, [we've got Rubocop configured](https://github.com/ministryofjustice/prison-visits/blob/c2d8e43b75fa2d284a9b3ecbb8dded308930618f/.default_rubocop.yml#L12-L14) for [Weirich-style blocks](http://devblog.avdi.org/2011/07/26/the-procedurefunction-block-convention-in-ruby/), i.e. `{...}` for blocks that return a value, and `do ... end` for imperative code.

tatyree: As an old cat if mine would have said: "Hmph".  It burns my eyes, Monk.  It burns my eyes...

However, I also confess that I secretly like it and shall change styles, myself :)

threedaymonk: The indentation here is kind of ludicrous, though, isn't it? I think I'll change it to:

```ruby
let(:data) {
  path = File.expand_path('../../fixtures/deserialization_data.yml', __FILE__)
  YAML.load(File.read(path))
}
```

tatyree: That looks cleaner and isn't excessively long.

config/initializers/encryptor.rb
```diff
@@ -1,6 +1,2 @@
-require 'zmarshal'
-
-key_length = 88
-key = ENV.fetch('MESSAGE_ENCRYPTOR_SECRET_KEY') { 'a' * key_length }
-MESSAGE_ENCRYPTOR = ActiveSupport::MessageEncryptor.new \
-  Base64.decode64(key), serializer: ZMarshal
+key = ENV.fetch('MESSAGE_ENCRYPTOR_SECRET_KEY') { 'a' * 88 }
```
tatyree: Is the { 'a' * 88 } an implicit pack?  I've never seen this syntax before...

app/models/deferred/visitor.rb
```diff
@@ -0,0 +1,4 @@
+module Deferred
+  class Visitor < ::Visitor
+  end
```
SteveMarshall: I wonder if it's worth having a blaring warning to never do anything in this class here?

app/services/compressed_marshaller.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def serialize(obj)
+    JSON.generate(obj)
+  end
+
+  def deserialize(str)
+    JSON.parse(str)
+  rescue JSON::ParserError
+    Rails.logger.info 'Deserializing legacy token'
```
SteveMarshall: We should probably test this on preprod, to make sure we can graph it in Kibana or something.

threedaymonk:
I wonder if it's worth adding some specs for the current serialised data. It won't prove anything at the moment, but might prevent future collaborators from repeating the problem we saw the other day.

app/models/deferred/visitor.rb
tatyree: I'd vote for that.

threedaymonk: Yes, it probably is!

SteveMarshall:
Yeah, it might be nice to have some specs for the current data.

The other thing that might be nice, maybe, is breaking the long cipher texts in the specs over multiple lines? Of course, that might be dangerous, so I'm willing to be told I'm wrong.

threedaymonk:
@SteveMarshall That should be possible: it's just base64-encoded, so `%w[ aaaa bbbb cccc ]`.join('') ought to work.

app/models/deferred/visitor.rb
threedaymonk: I'm tempted to write something like the [10,000 year nuclear waste warning](http://www.wipp.energy.gov/picsprog/articles/wipp%20exhibit%20message%20to%2012,000%20a_d.htm):

> This place is not a place of honor.
> No highly esteemed deed is commemorated here.
> Nothing valued is here.
> This place is a message and part of a system of messages.
> Pay attention to it!
> Sending this message was important to us.
> We considered ourselves to be a powerful culture.

tatyree: It appeals to my literary side.  However, I think it might require too much interpretation on the part of the average developer...

My suggestion would be to give it an initialiser that raises an exception if anyone tries to define instance methods at this level, but otherwise passes them down (up?) the chain.  That might be overcomplicating it, though.

threedaymonk: Possibly, something as simple as removing the `.new` and `#initialize` methods might be enough - I'm sure that `Marshal.load` bypasses that.

threedaymonk: Undefining `initialize` works, but spams the test output with warnings. I'm inclined to leave it with a comment.

tatyree: My thinking was more to make it explicit to anyone new coming to that particular class that WRITE THEE NO METHODS HERE: 

Something like: 

```
module Deferred 
  class Visitor < ::Visitor 
    
    class WriteTheeNoMethodsHere < Error 
       "This is a DANGEROUS place.  Leave here now..." 
    end

    def initialize 
       raise WriteTheeNoMethodsHere if self.instance_methods(false)
    end
end
```

threedaymonk:
Tested on preprod for both visitor and staff sides in two scenarios:

1. With booking emails sent by the currently-deployed production branch, after deploying this branch.
2. With an entirely new booking.

app/services/compressed_marshaller.rb
threedaymonk: Tested as working.

screenshot-2015-10-01t13 01 44z

PR #337, opened 2015-09-28
Remove browserstack rake task and integration

tatyree:
@SteveMarshall thinks this was the original attempt at developing smoke tests. It
was commented out in .circle.yml with an instruction to uncomment it
when it was made to work.  I could not get it to run at all locally and
the jenkins job it was supposedly powering had not succeeded once in
over 300 runs since early November 2014.

I am now going to add a ticket to the backlog to re-implement it using a
more ruby-like approach (i.e. not spawning external java classes to
proxy communication to Browserstack by overriding the port assigments of
a webbrick server, &tc. &tc.)


SteveMarshall:
LGTM! Got some conflicts, though…

tatyree:
@SteveMarshall--sorted.

spec/poltergeist_helper.rb
```diff
@@ -0,0 +1,7 @@
+require 'rails_helper'
```
threedaymonk: Is there any reason not to move this into `rails_helper.rb`? It would be more in line with other projects I've worked on.

tatyree: My thinking at the time was that it had very limited scope and that we might kill all those specs, entirely with the introduction of the smoke tests.

threedaymonk: You're right about the scope. I don't think that the smoke tests invalidate any of those specs, though.

PR #338, opened 2015-09-28
Disable key- or IP-only restrictions

SteveMarshall:
Wherever we're restricting access to an endpoint by IP, we want to be able
to access that using the auth key, too. That way, people who're allowed to
perform restricted actions don't have to always be on the network we
expect!

This also removes our bespoke secure_compare, because there's no real
reason for it to exist.

[fixes #104319098]


app/controllers/application_controller.rb
```diff
@@ -3,38 +3,32 @@ class ApplicationController < ActionController::Base
...
-  def self.permit_only_with_key
-    before_filter :reject_without_key!
+    before_filter :reject_without_key_or_trusted_ip!
   end
 
-  def reject!
-    raise ActionController::RoutingError.new('Go away')
+  def reject_without_key_or_trusted_ip!
+    unless valid_auth_key? || permitted_ip?
+      raise ActionController::RoutingError.new('Go away')
```
tatyree: Shouldn't this be in translations?  Also, isn't 'Go away' dangerously close to violating the "don't be cute" maxim?

SteveMarshall: Yes, on both counts. I couldn't think what to replace this crappy copy with. Probably this should be a bog-standard 404.

tatyree: Yeah, a 404 seems better.  The 'Go away' also suggests there may be something of interest there.

SteveMarshall: Hm. Do we think `AbstractController::ActionNotFound` or `AbstractController::RoutingError` is the best fit for this? (I'll remove the `Go away`, of course, but I figure i may as well use the right exception while I'm here…)

tatyree: To me `AbstractController::RoutingError` seems conceptually cleaner in that it is actually an attempt to follow a route they aren't allowed access to.

SteveMarshall: Fixed in e0a0997. Lemme know what you think and I'll rebase.

SteveMarshall:
Let's not merge this (or any of the others) till @threedaymonk is done with #336.

app/controllers/application_controller.rb
tatyree: LGTM.  Ship it! (when @threedaymonk has merged)

app/controllers/application_controller.rb
```diff
@@ -3,38 +3,32 @@ class ApplicationController < ActionController::Base
   before_filter :add_extra_sentry_metadata
   protect_from_forgery with: :exception
 
-  def self.permit_only_from_prisons
-    before_filter :reject_untrusted_ips!
-  end
-
   def self.permit_only_from_prisons_or_with_key
```
threedaymonk: This is literally what it does, but perhaps the important intention is slightly different: `permit_only_trusted_users` or similar (where trusted means either on a trusted IP or authenticated).

app/controllers/webhooks_controller.rb
```diff
@@ -69,7 +69,11 @@ def email_params
   end
 
   def authorized?
-    params[:auth].present? && params[:auth].secure_compare(ENV['WEBHOOK_AUTH_KEY'])
+    params[:auth].present? &&
```
threedaymonk: Is it important to check whether `params[:auth]` is supplied but an empty string? If we're only checking so that we don't blow up on nil, perhaps `params.key?(:auth)` is better.

app/controllers/webhooks_controller.rb
```diff
@@ -69,7 +69,11 @@ def email_params
...
+      ActiveSupport::SecurityUtils.secure_compare(
+        params[:auth],
```
threedaymonk: According to [the documentation of `secure_compare`](http://api.rubyonrails.org/classes/ActiveSupport/SecurityUtils.html#method-c-secure_compare):

> This should not be used on variable length plaintext strings because it could leak length info via timing attacks.

That implies that we should do something to fix the length beforehand. I _think_ this would work:

```ruby
length = ENV.fetch('WEBHOOK_AUTH_KEY').length
fixed_length_param = (params.fetch(:auth) + ' ' * length).slice(0, length)
```

… but I don't know whether that has a fixed time cost.

SteveMarshall: Good catch. That implementation would also leave us open to inexact matches where `WEBHOOK_AUTH_KEY` is shorter than the `auth` param, wouldn't it?

That said, the first line of `secure_compare` does this, which I'd have thought covers us (or is `bytesize` comparison not fixed time?):

```ruby
return false unless a.bytesize == b.bytesize
```

threedaymonk: I think the problem is that this returns quicker when there's a length mismatch, which allows an attacker to determine the key length.

config/initializers/string.rb
```diff
@@ -1,11 +0,0 @@
-class String
```
dan-palmer: 😐

threedaymonk:
Looks good.

PR #340, opened 2015-09-30
Make validation account for /prisoner in preprod

tatyree:
https://preprod-prisonvisits.dsd.io 301s to production.  You can only
view the preprod site if you hit it on
https://preprod-prisonvisits.dsd.io/prisoner.  This breaks the smoke
test. I thought it would be enough to check that the URL had the correct
path in it somewhere.


SteveMarshall:
To clarify: this is a safeguard against someone inadvertently starting smoke-testing against `/` instead of `/prisoner`, right?

tatyree:
Not exactly.  The smoke test was hitting /prisoner.  This caused the first test to fail on preprod, which responds on /prisoner/prisoner.

smoke_test/steps/prisoner_page.rb
```diff
@@ -1,11 +1,11 @@
 module SmokeTest
   module Steps
     class PrisonerPage < BaseStep
-      PAGE_PATH = '/prisoner'
+      PAGE_PATH = %r{/prisoner$}
```
dan-palmer: I don't think this is correct - should expect the url to be something like: `https://www.prisonvisits.service.gov.uk/prisoner`

tatyree:
Not needed.  `/prisoner` was being accidentally included in the host environment variable.

PR #341, opened 2015-10-01
Make session timeout longer in development

EGiataganas:
Now the session timeout is 20 minutes for staging and production,
but for the development is 24 hours so its less aggressive now.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -42,5 +42,6 @@ module PrisonVisits2
     config.autoload_paths += %w(lib app/mailers/concerns)
     config.assets.paths << Rails.root.join('vendor', 'assets', 'moj.slot-picker', 'dist', 'stylesheets')
     config.middleware.delete 'ActiveRecord::QueryCache'
+    config.session_expire_after = 20.minutes
   end
 end
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -3,7 +3,7 @@ SERVICE_DOMAIN = (service_url) ? URI.parse(service_url).host : nil
 
 PrisonVisits2::Application.config.session_store :cookie_store,
   key: 'pvbs',
-  expire_after: 20.minutes,
+  expire_after: Rails.configuration.session_expire_after,
   httponly: true,
   domain: SERVICE_DOMAIN,
   secure: Rails.env.production?
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -5,6 +5,5 @@ PrisonVisits2::Application.config.session_store :cookie_store,
   key: 'pvbs',
   expire_after: 20.minutes,
   httponly: true,
-  max_age: 20.minutes,
   domain: SERVICE_DOMAIN,
   secure: Rails.env.production?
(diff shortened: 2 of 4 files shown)


config/initializers/session_store.rb
```diff
@@ -1,9 +1,16 @@
 service_url = ENV['SERVICE_URL']
 SERVICE_DOMAIN = (service_url) ? URI.parse(service_url).host : nil
+timeout_time = ENV['TIMEOUT_TIME']
```
tatyree: `TIMEOUT_TIME` is a bit redundant.  I'd prefer just `timeout` for the variable name and something more specific for the environment variable, `PVB_SESSION_TIMEOUT`, perhaps.

config/initializers/session_store.rb
```diff
@@ -1,9 +1,16 @@
...
+
+if Rails.env.developement?
```
SteveMarshall: We shouldn't be switching on `Rails.env` values; this should entirely be handled by the environment variables, possibly with a default.

config/initializers/session_store.rb
```diff
@@ -1,9 +1,16 @@
...
+  timeout_time
```
SteveMarshall: This line does nothing; should it be trying to convert `timeout_time` into minutes?

config/initializers/session_store.rb
```diff
@@ -1,9 +1,16 @@
...
+else
+  timeout_time = 20.minutes
```
SteveMarshall: This default could be handled by `ENV.fetch('TIMEOUT_TIME', 20)` or similar.

config/initializers/session_store.rb
SteveMarshall: Might be better if the environment variable was `SESSION_TIMEOUT`, maybe?

tatyree: @SteveMarshall I've currently got three apps on my machine that respond to SESSION_TIMEOUT, each with different requirements.  I'd prefer we either set it in a PVB specific dotenv file or namespace it to the app.

config/initializers/session_store.rb
```diff
@@ -3,8 +3,7 @@
 
 PrisonVisits2::Application.config.session_store :cookie_store,
   key: 'pvbs',
-  expire_after: 20.minutes,
+  expire_after: Rails.configuration.session_expires_after,
   httponly: true,
-  max_age: 20.minutes,
```
threedaymonk: We know that we don't need this because it has exactly the same function as `:expire_after`: Rails's [`CookieStore`](https://github.com/rails/rails/blob/3ac3760c69e6e6914c5ddae138856b3c82ac0f20/actionpack/lib/action_dispatch/middleware/session/cookie_store.rb#L60) is a subclass of [`Rack::Session::Abstract::ID`](https://github.com/rack/rack/blob/028438ffffd95ce1f6197d38c04fa5ea6a034a85/lib/rack/session/abstract/id.rb#L349-L350), in which they are treated identically.

config/application.rb
```diff
@@ -42,5 +42,6 @@ class Application < Rails::Application
     config.autoload_paths += %w(lib app/mailers/concerns)
     config.assets.paths << Rails.root.join('vendor', 'assets', 'moj.slot-picker', 'dist', 'stylesheets')
     config.middleware.delete 'ActiveRecord::QueryCache'
+    config.session_expires_after = 20.minutes
```
threedaymonk: I think we should rename this to `sessions_expire_after` (no s) so that it's consistent with the cookie parameter (`:expires_after`).

SteveMarshall: :+1:

SteveMarshall:
@EGiataganas Have you tested this on preprod?

EGiataganas:
@SteveMarshall Tested on pre-prod and after 20 minutes you will have redirect to session expire page

PR #342, opened 2015-10-01
Make `/` only route to start page on prod

EGiataganas:
We change route so it doesn't redirect to the live environment in the
dev environment.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -68,5 +68,5 @@ PrisonVisits2::Application.routes.draw do
   # Legacy URLs
   get "/prisoner-details", to: redirect("/prisoner")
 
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -68,5 +68,9 @@
   # Legacy URLs
   get "/prisoner-details", to: redirect("/prisoner")
 
-  get "/", to: redirect(ENV["GOVUK_START_PAGE"] || "https://www.gov.uk/prison-visits")
+  if Rails.env.production?
+    get "/", to: redirect(ENV["GOVUK_START_PAGE"] || "https://www.gov.uk/prison-visits")
```
SteveMarshall: I think we can probably remove the `ENV` here and maybe move the start page URL to the Rails config? @threedaymonk / @tatyree any preferences?

EGiataganas:
@SteveMarshall Tested on pre-prod

PR #344, opened 2015-10-02
Change timeouts to account for sidekiq queue

tatyree:
This has changed the amount of time it takes for an email to get through
the system, resulting in smoke test timeouts.  This change should fix
that.


threedaymonk:
This seems reasonable.

PR #348, opened 2015-10-06
Allow simplecov to be disable by setting envvar

tatyree:
The tests are slow enough without having to wait the same amount of time
again on a coverage report that is duplicated in code climate.  I find
simplecov reporting useful periodically, but get particularly annoyed
with it when I'm iterating on a single spec and it takes three times as
long as the spec, itself.

This allows you to `export DISABLE_SIMPLECOV=true` to turn it off.


threedaymonk:
I wonder if it might be possible to do this automatically using RSpec's `files_to_run.one?`. I'll have a look now.

PR #354, opened 2015-10-09
Fix incorrect input validation on age fields

threedaymonk:
The maximum value for date input fields (enforced in modern browsers) was incorrectly calculated as 120 years ago from today, and not the end of the year. This meant that, for example, on 9 October, dates with a day greater than 9 or a month greater than 10 could not be entered.


threedaymonk:
Tested on preprod.

PR #355, opened 2015-10-09
Fix date method

EGiataganas:


app/mailers/prison_mailer.rb
```diff
@@ -107,4 +107,14 @@ def smoke_test?
...
+
+  def first_date
+    Date.parse(@visit.slots.first.date)
+  end
+
+  def request_date
+    first_date.strftime('%A %e %B')
+  end
+
+  alias_method :receipt_date, :request_date
```
SteveMarshall: Should these be the same? @tatyree, isn't this what you were talking about this morning?

app/mailers/prison_mailer.rb
```diff
@@ -107,4 +107,8 @@ def smoke_test?
   def visitors_email_address
     visit.visitors.first.email
   end
+
+  def formatted_date
+    Date.parse(@visit.slots.first.date).strftime('%A %e %B')
```
threedaymonk: I think I'd go one further and make this a named format (see http://api.rubyonrails.org/classes/DateTime.html#method-i-to_formatted_s and https://github.com/ministryofjustice/peoplefinder/blob/master/config/locales/en.yml#L22-L25 for examples).

app/mailers/prison_mailer.rb
```diff
@@ -107,4 +107,8 @@ def smoke_test?
   def visitors_email_address
     visit.visitors.first.email
   end
+
+  def formatted_date
+    l(Date.parse(@visit.slots.first.date), format: :short)
+  end
```
threedaymonk: I think it might be clearer to future readers if we used `I18n.l` here.

I also think we could rename the method to something like `formatted_first_slot_date` to explain what the date is.

app/mailers/visitor_mailer.rb
```diff
@@ -73,13 +67,11 @@ def first_date
...
+  def slot_date
+    Date.parse(@slot.date)
   end
 
-  alias_method :rejection_date, :receipt_date
-
-  def confirmation_date
-    Date.parse(@slot.date).strftime('%e %B %Y').gsub(/^ /,'')
+  def format_date(date)
+    date.strftime('%e %B %Y').gsub(/^ /,'')
```
threedaymonk: This could also be a named format. Note that instead of `%e` with `gsub` to remove the padding, [we can use `%-d`](http://ruby-doc.org/core-2.0.0/Time.html#method-i-strftime):

```
%d - Day of the month, zero-padded (01..31)
        %-d  no-padded (1..31)
%e - Day of the month, blank-padded ( 1..31)
```

threedaymonk: It's also worth looking into why this date has a year and the one above doesn't. Perhaps they could both be the same.

config/locales/en.yml
```diff
@@ -2,6 +2,7 @@ en:
   date:
     formats:
       default: "%d/%m/%Y"
+      short: "%A %e %B"
```
threedaymonk: Instead of `short`, how about something like `day_and_date`?

app/mailers/prison_mailer.rb
EGiataganas: @threedaymonk yeah sure :+1:

config/locales/en.yml
EGiataganas: @threedaymonk yes much better, tbh I couldn't think any different name :+1:

threedaymonk:
Looks good to me.

As an aside, this raises the broader question of whether our email titles are misleading: because you can apply for multiple slots on multiple days, in order of preference, things like "Visit request for %{full_name} on %{request_date}" aren't really accurate, and the rejection is even worse (e.g. you apply for Tuesday, Wednesday, and Thursday, and the email just says "your visit for Tuesday could not be booked"). I assume that this is a vestige of a time before second and third preferences.

@leighmoney what do you think?

PR #356, opened 2015-10-09
Eliminate avoidable 500 errors

threedaymonk:
Previously, admin staff following a link in which the `state` parameter (the serialised application state) was corrupt received a helpful error page, but members of the pupblic just saw a generic error page, and we logged a 500 error. Most 500 errors logged are due to this.

This changes brings both pages to parity, and makes the behaviour the same for any point in the application that deals with serialised data.


dan-palmer:
LGTM :+1:

PR #357, opened 2015-10-12
 Change the way prison data is accessed.

tatyree:
Previously, it had been loaded into Rails.configuration as a hash and accessed as a hash.  This branch changes that such that all access now *must* be done using the Prison model and its instances.

I'll go ahead and start rebasing master in while it is being reviewed.


app/controllers/static_controller.rb
```diff
@@ -6,8 +6,8 @@ def prison_emails
       format.csv do
         csv_string = CSV.generate do |csv|
           csv << ['name', 'email']
-          Rails.configuration.prison_data.each_pair do |prison_name, data|
-            csv << [prison_name, data[:email]] if data[:enabled]
+          Prison.enabled do |prison|
```
threedaymonk: The lack of `each` here feels a bit confusing.

app/helpers/visit_helper.rb
```diff
@@ -18,44 +16,45 @@ def prison_name
...
 
   def prison_data(source=visit)
-    Rails.configuration.prison_data.fetch(source.prisoner.prison_name.to_s)
+    prison_name = source.is_a?(String) ? source : source.prisoner.prison_name.to_s
+    Prison.find(prison_name)
   end
 
   def prison_phone
-    prison_data['phone']
+    prison_data.phone
```
threedaymonk: Some of these could be implemented with delegation:

```ruby
delegate :phone, :email, to: prison_data, prefix: :prison
```

ought to do it.

app/helpers/visit_helper.rb
```diff
@@ -18,44 +16,45 @@ def prison_name
...
 
   def adult_age
-    prison_data['adult_age'] || 18
+    prison_data.adult_age || 18
   end
 
   def prison_url(visit)
-    data = prison_data(visit)
-    slug = data['finder_slug'] ? data['finder_slug'] : visit.prisoner.prison_name.parameterize
+    return unless prison_data
```
threedaymonk: Why do we need to check for `nil` here , when we don't for any of the other attributes?

app/models/prison.rb
```diff
@@ -1,32 +1,62 @@
 class Prison
+  attr_accessor :address, :adult_age, :booking_window, :canned_responses,
+    :email, :enabled, :finder_slug, :lead_days,
+    :name, :nomis_id, :phone, :reason,
+    :slot_anomalies, :slots, :unbookable,
+    :works_weekends
```
threedaymonk: This seems to be wrapped more than necessary.

app/models/prison.rb
```diff
@@ -1,32 +1,62 @@
...
+
   DEFAULT_LEAD_DAYS = 3.freeze
   DEFAULT_BOOKING_WINDOW = 28.freeze
```
threedaymonk: `freeze` on an integer doesn't do anything.

app/models/prison.rb
```diff
@@ -1,32 +1,62 @@
...
       raise PrisonNotFound, "Can't find prison #{prison_name.inspect}"
     end
-    new(prison_name, prison_hash)
+    prison
+  end
+
+  def initialize(opts = {})
+    opts.with_indifferent_access.map do |k, v|
+      instance_variable_set("@#{k}", v)
+    end
```
threedaymonk: This behaviour comes for free with Virtus, which we're using already.

app/models/prison.rb
```diff
@@ -1,32 +1,62 @@
...
   end
 
-  attr_reader :name
+  def self.names
+    all.map(&:name).sort
+  end
 
-  delegate :fetch, to: :@prison_hash
+  def self.nomis_ids
+    all.map(&:nomis_id).reject(&:nil?).sort
```
threedaymonk: `reject(&:nil?)` can be simplified to `compact`.

app/models/prison.rb
```diff
@@ -1,32 +1,62 @@
...
+  end
 
   def unbookable_dates
-    (fetch(:unbookable) { Array.new }).to_set
+    (unbookable || Array.new).to_set
   end
 
   def visiting_slots
-    fetch(:slots)
+    @visiting_slots ||= slots
```
threedaymonk: if this method is now just an alias, why memoise. (Or, why memoise here, and not there?)

app/models/prison.rb
```diff
@@ -1,32 +1,62 @@
...
+    all.map(&:nomis_id).reject(&:nil?).sort
+  end
+
+  def enabled?
+    enabled
+  end
 
   def unbookable_dates
-    (fetch(:unbookable) { Array.new }).to_set
+    (unbookable || Array.new).to_set
```
threedaymonk: `Array.new` is better as just `[]`

app/models/prisoner.rb
```diff
@@ -6,13 +6,19 @@ class Prisoner
   attribute :prison_name, String
 
   validates_format_of :number, with: /\A[a-z]\d{4}[a-z]{2}\z/i, message: "must be a valid prisoner number" # eg a1234aa
-  validates_inclusion_of :prison_name, in: Rails.configuration.prison_data.keys, message: "must be chosen"
+  validates_inclusion_of :prison_name, in: Prison.names, message: "must be chosen"
   validate :prison_in_service
 
   def prison_in_service
-    if !self.prison_name.blank? && !Rails.configuration.prison_data[self.prison_name]['enabled']
+    unless prison && prison.try(:enabled)
```
threedaymonk: `&&` and `try` is duplication, and it would be nice to use the `enabled?` method you defined above.

```ruby
unless prison.try(:enabled?)
```

or

```ruby
unless prison && prison.enabled?
```

(my preference).

app/mailers/concerns/addresses.rb
```diff
@@ -2,7 +2,7 @@ module Addresses
   include ActiveSupport::Concern
 
   def prison_mailbox_email
-    Rails.configuration.prison_data[@visit.prisoner.prison_name]['email']
+    Prison.find(@visit.prisoner.prison_name).email
```
threedaymonk: couldn't this just be `@visit.prisoner.prison.email`? It would be even nicer if we could delegate through: `@visit.prison.email` or `@visit.prison_email` perhaps.

app/models/visit.rb
```diff
@@ -27,7 +27,7 @@ def validate_amount_of_adults
   end
 
   def adult_age
-    AgeValidator.new(Rails.configuration.prison_data[prisoner.prison_name]).adult_age
+    AgeValidator.new(prisoner.prison_name).adult_age
```
threedaymonk: I feel like this should probably take the prison as a parameter, not do its own lookup.

app/views/confirmations/new.html.haml
```diff
@@ -1,7 +1,7 @@
 - content_for :header do
   Process a visit request
 
-- if prison_data(booked_visit)['canned_responses']
+- if Prison.find(booked_visit.prisoner.prison_name).canned_responses
```
threedaymonk: Would be nice as `booked_visit.prison.canned_responses`.

config/prison_data.yml
```diff
@@ -1641,20 +1639,10 @@ High Down:
...
-  - 2015-06-10
-  - 2015-07-08
-  - 2015-08-12
-  - 2015-09-09
+  - 2015-09-16
   - 2015-10-14
-  - 2015-11-11
-  - 2015-12-09
+  - 2015-11-18
+  - 2015-12-16
```
threedaymonk: Check all these missing values.

config/prison_data.yml
```diff
@@ -1666,13 +1654,8 @@ High Down:
     - 1400-1600
     wed:
     - 1400-1600
-  unbookable:
-  - 2015-09-16
-  - 2015-10-14
-  - 2015-11-18
-  - 2015-12-16
```
threedaymonk: Check all these missing values.

config/prison_data.yml
```diff
@@ -3356,92 +3339,90 @@ Channings Wood:
...
     - 1400-1600
-Acklington:
-  nomis_id: AKI
+  name: Channings Wood
+- nomis_id: AKI
   enabled: false
   finder_slug: acklington2
-Albany:
-  nomis_id: ALI
-  enabled: false
```
threedaymonk: Where did Albany go?

lib/age_validator.rb
```diff
@@ -1,9 +1,11 @@
 class AgeValidator < ActiveModel::Validator
   DEFAULT_ADULT_AGE = 18
 
-  def initialize(prison_config)
-    @config = prison_config || (raise ArgumentError)
-    super
+  attr_accessor :adult_age
+
+  def initialize(prison_name)
+    prison = Prison.find(prison_name, true)
```
threedaymonk: As mentioned above, I feel that this should take the prison directly, not look it up.

lib/age_validator.rb
```diff
@@ -1,9 +1,11 @@
...
+    @adult_age = prison.adult_age || DEFAULT_ADULT_AGE
```
threedaymonk: Could prison be responsible for the default adult age?

lib/metrics_logger.rb
```diff
@@ -1,6 +1,6 @@
 class MetricsLogger
   def record_visit_request(visit)
-    nomis_id = Rails.configuration.prison_data[visit.prisoner.prison_name]['nomis_id']
+    nomis_id = Prison.find(visit.prisoner.prison_name).nomis_id
```
threedaymonk: can this be `visit.prison.nomis_id`?

app/helpers/visit_helper.rb
tatyree: There's at least one spec (a view spec, I think) that breaks if this doesn't check for nil.   Might be better to just try(:finder_slug) instead.

app/models/prison.rb
dan-palmer: Think I'm probably to blame for this.. :cry: total dufus. Integers & Floats are frozen by default.

config/environments/development.rb
```diff
@@ -15,4 +15,8 @@
   config.smoke_test_email_local_part = ENV['SMOKE_TEST_EMAIL_LOCAL_PART']
   config.smoke_test_email_domain = ENV['SMOKE_TEST_EMAIL_DOMAIN']
   config.session_expire_after = 24.hours
+
+  config.prison_data = YAML.load_file(
```
dan-palmer: Couldn't this just be in `application.rb` as it's not environment specific?

SteveMarshall:
A couple of general questions:

- The move of the (prison) `name` to the end of each object makes the diff for, and probably editing of, `prison_data.yml` confusing (in 2a32fa3). It might be clearer what's going on (and easier for people to edit) if the names were still directly beside the NOMIS IDs?
- Have you tested what might happen with bookings made before applying your changes? Will they continue to work, or will this break them? Possibly worth looking at what @threedaymonk did in #336 (in particular, `spec/features/deserialization_spec.rb`), and seeing if there's anything there that we might want to extend for this change?

config/environments/development.rb
SteveMarshall: :+1:

app/helpers/visit_helper.rb
```diff
@@ -18,44 +16,45 @@ def prison_name
   end
 
   def prison_names
-    Rails.configuration.prison_data.keys.sort
+    Prison.names
   end
 
   def prison_data(source=visit)
-    Rails.configuration.prison_data.fetch(source.prisoner.prison_name.to_s)
+    prison_name = source.is_a?(String) ? source : source.prisoner.prison_name.to_s
```
dan-palmer: A couple of things stand out here - whether they are addressed in this PR or not is a different matter.

Firstly, minor niggle, we are no longer returning a hash, so calling it data I feel takes away from it being an object in its own right now & is the start of it reading a little inconsistent.. sometimes we have prisons sometimes we have prison_data objects.

I also seem to recall seeing a number of similar methods that are auto seeded, and can take different inputs (which in most cases is a lazy way of defining a common interface) in this case `String` or `Visit`.

I think we can clear this up a little.. We can fall into two camps here either always pass the object to the method or accept that it is always available to us - `prison_data` seems to be commonly called without arguments so to make it consistent we can go with the latter & do away with having this method take any argument as `visit` is a global helper method defined in `ApplicationController` & is always available.

Doing a grep for this method I can find only 2 instances of it being called with args.

```bash
$ git grep "prison_data(" | grep -v def

app/helpers/visit_helper.rb:    slug = prison_data(visit).finder_slug || visit.prisoner.prison_name.parameterize
app/helpers/visit_helper.rb:    nomis_id = prison_data(prison_name).nomis_id
```

For the `slug` assignment we don't need to pass anything as its auto seeded with the visit anyway/as mentioned earlier visit is always available within the scope of this method.

For the `nomis_id` assignment used in the method `prison_specific_id_requirements` we can see that it is used in the `app/views/content/_id_requirements.text.erb` and `app/views/content/_id_requirements.html.erb` & what a surprise is being passed the sessions prison name via the prisoner in order for lookup in the prison hash.

And then for the actual contents of this method would it make more sense for a visit to be asked what prison is assigned to it and return said object or `nil`?

That way we could do something like..

```ruby
# please rename me to prison :sob:

def prison_data
  visit.prisoner.prison
end

# which could be refactored eventually to something like..

delegate :prison, to: :visit
```

config/prison_data.yml
tatyree: It's now 'Isle of Wight - Albany'

app/helpers/visit_helper.rb
dan-palmer: Wasn't ranting btw ^ :laughing:

config/environments/development.rb
tatyree: @dan-palmer No, because models aren't yet initialised when application.rb loads.  Works fine in an initialiser, though.

tatyree:
@SteveMarshall have a look at the new, sorted data structure and see if that helps.   I'll discuss testing and possible breakage with @threedaymonk tomorrow morning.

app/helpers/visit_helper.rb
tatyree: I've changed prison_data in the initializer and Prison model.  I'll do it in the `visit_helper.rb` tomorrow before I rebase against master.

app/helpers/visit_helper.rb
```diff
@@ -8,57 +8,43 @@ def visiting_slots
...
   end
 
   def prison_names
-    Rails.configuration.prison_data.keys.sort
+    Prison.names
   end
 
-  def prison_data(source=visit)
-    Rails.configuration.prison_data.fetch(source.prisoner.prison_name.to_s)
+  def prison(source=visit)
```
dan-palmer: Is this ever called with a `String` argument?

tatyree: It was being, but I think I removed that.  I'll check.

dan-palmer: I might be wrong.. but I think `prison` could just be delegated to 'visit'

tatyree: Yes. It can be. Doing it now.

app/helpers/visit_helper.rb
```diff
@@ -81,17 +67,14 @@ def slots_for_day(day)
   end
 
   def when_to_expect_reply
-    prison = Prison.find(visit.prisoner.prison_name)
-    schedule = PrisonSchedule.new(prison)
+    schedule = PrisonSchedule.new(visit.prison)
```
dan-palmer: This could just be:

```ruby
schedule = PrisonSchedule.new(prison)
```

app/models/prison.rb
```diff
@@ -1,32 +1,77 @@
...
+  end
 
   def unbookable_dates
-    (fetch(:unbookable) { Array.new }).to_set
+    (unbookable || []).to_set
   end
 
   def visiting_slots
-    fetch(:slots)
+    @visiting_slots = slots
```
dan-palmer: Couldn't this just be:

```
def visiting_slots
  slots
end

# or

alias_method :visiting_slots, :slots
```

app/models/prison.rb
```diff
@@ -34,19 +79,15 @@ def visiting_slot_days
...
-    fetch(:lead_days, DEFAULT_LEAD_DAYS)
-  end
-
-  def booking_window
-    fetch(:booking_window, DEFAULT_BOOKING_WINDOW)
+    lead_days
   end
 
   def works_weekends?
-    fetch(:works_weekends, false)
```
dan-palmer: If we have allowed Virtus to set a default value we don't need the fallback here.

app/models/prison.rb
```diff
@@ -1,32 +1,77 @@
 class Prison
-  DEFAULT_LEAD_DAYS = 3.freeze
-  DEFAULT_BOOKING_WINDOW = 28.freeze
+  include Virtus.model
+
+  DEFAULT_ADULT_AGE =       18
+  DEFAULT_BOOKING_WINDOW =  28
+  DEFAULT_LEAD_DAYS =       3
+
+  attribute :address
```
dan-palmer: Quite like Virtus & feel it works well here :+1:

tatyree: Agreed.

app/controllers/metrics_controller.rb
```diff
@@ -81,7 +81,7 @@ def fortnightly_range
   end
 
   def prison_param
-    prison = params[:prison]
+    (prison = params[:prison]) && Prison.nomis_ids.include?(prison) && prison
```
dan-palmer: :fearful: This is a bit confusing

dan-palmer: Line 85 shouldn't be there right?

app/models/prison.rb
```diff
@@ -1,32 +1,77 @@
...
-      raise PrisonNotFound, "Can't find prison #{prison_name.inspect}"
+  def self.find(name_or_nomis, raise_error = false)
+    prison = all.detect { |p|
+      p.name == name_or_nomis || p.nomis_id == name_or_nomis
+    }
+    if raise_error && prison.nil?
+      raise PrisonNotFound, "Can't find prison #{name_or_nomis.inspect}"
     end
-    new(prison_name, prison_hash)
+    prison
```
dan-palmer: Should we duplicate the prison at this point?

```ruby
prison1 = Prison.find 'Wandsworth'
prison1.unbookable                          # => list of dates..
prison1.unbookable = []
prison2 = Prison.find 'Wandsworth'
prison2.unbookable                          # => [] empty list :sadface:
```

dan-palmer: Is the raise_error used anywhere?

If it is.. might it be nice to have a bang alternative that handles the errors?

app/controllers/metrics_controller.rb
tatyree: L85 was a merge error.  Yes, it is confusing, but so is the original...

app/models/prison.rb
```diff
@@ -1,32 +1,77 @@
...
+  end
+
+  def initialize(opts = {})
+    self.attributes = attributes.merge(opts.symbolize_keys)
+  end
+
+  def self.create(opts = {})
+    prison = new(opts)
```
dan-palmer: Perhaps `tap` it?

```ruby
new(opts).tap { |prison| Rails.configuration.prisons << prison }
```

app/controllers/metrics_controller.rb
dan-palmer: :laughing: called `prison_param` but checks if its in the nomis list :suspect:

app/views/confirmations/new.html.haml
```diff
@@ -1,7 +1,7 @@
 - content_for :header do
   Process a visit request
 
-- if prison_data(booked_visit)['canned_responses']
+- if booked_visit.prison_canned_responses
```
dan-palmer: Such an improvement :star2:

app/controllers/metrics_controller.rb
tatyree: True dat.  I'm a bit fried at this point :dizzy_face:

app/views/content/_id_requirements.html.erb
```diff
@@ -1 +1 @@
-<%= markdown(prison_specific_id_requirements(visit.prisoner.prison_name)) %>
+<%= markdown(prison_specific_id_requirements(visit.prison)) %>
```
dan-palmer: If `prison` is available via the helper now can we just do:

```ruby
<%= markdown(prison_specific_id_requirements(prison)) %>
```

Same would apply for the `text.er` of this template.

config/initializers/prisons.rb
```diff
@@ -0,0 +1,3 @@
+Rails.configuration.prisons = YAML.load_file(
```
dan-palmer: Such DRYing :smile:

lib/age_validator.rb
```diff
@@ -1,9 +1,8 @@
 class AgeValidator < ActiveModel::Validator
-  DEFAULT_ADULT_AGE = 18
```
dan-palmer: :+1:

config/initializers/prisons.rb
tatyree: I both :heart: and :frowning: when you're bored @dan-palmer

app/views/content/_id_requirements.html.erb
tatyree: Wow.  So many examples of this...Cleaning...

app/models/prison.rb
tatyree: @dan-palmer duplicating the found prison breaks a fair number of specs at this point.  I'm going to leave it for now and raise a ticket to look at it shortly.

app/helpers/visit_helper.rb
```diff
@@ -8,57 +8,38 @@ def visiting_slots
...
+    mail_to prison.email
   end
 
   def prison_postcode
-    prison_data['address'][-1]
-  end
-
-  def prison_slot_anomalies
-    prison_data['slot_anomalies']
+    prison.address[-1]
```
dan-palmer: Do you think `last` is preferable?

app/models/visit.rb
```diff
@@ -18,7 +18,10 @@ class Visit
...
   validate :validate_amount_of_adults, on: :visitors_set
 
-  delegate :prison_name, :prison_name=, to: :prisoner, allow_nil: true
+  delegate :prison, :prison_name, :prison_name=,
+    to: :prisoner, allow_nil: true
+  delegate :prison_email, :prison_canned_responses, :prison_nomis_id,
+    to: :prisoner
 
   def validate_amount_of_adults
     if visitors.none? { |v| v.age && v.age >= 18 }
```
dan-palmer: Regarding the magic number 18 here.. shouldn't this be using the prisons adult age?

threedaymonk: @dan-palmer you must be right. It's even starker [in context](https://github.com/ministryofjustice/prison-visits/blob/0e2a556899bc368e1abefbb7c28325f59d665404/app/models/visit.rb#L27-L32):

```ruby
    if visitors.none? { |v| v.age && v.age >= 18 }
      errors.add :visitors, :at_least_one_adult
    end
    if visitors.count { |v| v.age && v.age >= adult_age } > 3
      errors.add :visitors, :max_3_adults, adult_age: adult_age
    end
```

Those can't both be correct, surely?

tatyree: Actually it should be using its own method for this, I think...Although I think AgeValidator might be a candidate for deprecation now as it's only used to add a single, albeit complex, validation to Visit.   Can't see any reason why it needs to be it's own class any longer.

threedaymonk: And in fact, some kind of `adult?` method seems called for here:

```ruby
if visitors.none? { |v| adult?(v) }
```

app/helpers/visit_helper.rb
threedaymonk: This is actually the postcode, isn't it? Perhaps we could abstract it as such.

app/models/visit.rb
tatyree: There is one.  I just didn't notice it until now.

app/helpers/visit_helper.rb
threedaymonk: (on prison, I mean)

dan-palmer: Improvement++ :+1:

Prison data looks something like this:

```ruby
["Heathfield Rd", "PO Box 757", "SW18 3HS"]
```

app/models/visit.rb
```diff
@@ -18,20 +18,22 @@ class Visit
...
-    if visitors.count { |v| v.age && v.age >= adult_age } > 3
+    if visitors.count { |v| adult?(v) } > 3
       errors.add :visitors, :max_3_adults, adult_age: adult_age
     end
   end
 
   def adult_age
-    AgeValidator.new(Rails.configuration.prison_data[prisoner.prison_name]).
-      adult_age
+    AgeValidator.new(prison).adult_age
```
dan-palmer: Can't this just be:

```ruby
delegate :adult_age, to: :prison
```

dan-palmer:
Such work :tophat: @tatyree

app/helpers/visit_helper.rb
```diff
@@ -8,57 +8,38 @@ def visiting_slots
...
+    mail_to prison.email
   end
 
   def prison_postcode
-    prison_data['address'][-1]
-  end
-
-  def prison_slot_anomalies
-    prison_data['slot_anomalies']
+    prison.address.last
```
threedaymonk: Can we define this on `Prison` and delegate it?

app/helpers/visit_helper.rb
```diff
@@ -8,57 +8,38 @@ def visiting_slots
...
-    prison_data['adult_age'] || 18
+    safe_join(prison.address, glue)
   end
 
   def prison_url(visit)
-    data = prison_data(visit)
-    slug = data.fetch('finder_slug') { visit.prisoner.prison_name.parameterize }
-    ['http://www.justice.gov.uk/contacts/prison-finder', slug].join('/')
+    ['http://www.justice.gov.uk/contacts/prison-finder',
+     visit.prison_name.parameterize].join('/')
```
threedaymonk: I think this would be clearer as:

```ruby
[
  'http://www.justice.gov.uk/contacts/prison-finder',
  visit.prison_name.parameterize
].join('/')
```

**but** I also feel rather strongly that this URL should not be here.

app/helpers/visit_helper.rb
```diff
@@ -8,57 +8,38 @@ def visiting_slots
...
   end
 
   def prison_link(source=visit, link_text=nil)
-    link_text ||= "#{source.prisoner.prison_name.capitalize} prison"
+    unless link_text
+      link_text = "#{source.prison_name.capitalize} prison"
```
threedaymonk: Is there any circumstance in which a prison name needs `.capitalize`?

app/models/prison.rb
```diff
@@ -1,52 +1,88 @@
...
-  def initialize(name, prison_hash)
-    @name = name
-    @prison_hash = prison_hash.with_indifferent_access
+  def self.create(opts = {})
+    new(opts).tap { |p| Rails.configuration.prisons << p }
   end
 
-  attr_reader :name
+  def self.enabled
+    all.select{ |p| p.enabled == true }
```
threedaymonk: `all.select(&:enabled)` should work, no?

app/models/prison.rb
```diff
@@ -1,52 +1,88 @@
...
+  end
 
-  delegate :fetch, to: :@prison_hash
+  def self.find(name_or_nomis, raise_error = false)
```
threedaymonk: This `raise_error` parameter feels a bit unidiomatic to me.

app/models/prison.rb
```diff
@@ -1,52 +1,88 @@
...
+    if raise_error && prison.nil?
+      raise PrisonNotFound, "Can't find prison #{name_or_nomis.inspect}"
+    end
+    prison
+  end
 
-  def unbookable_dates
-    (fetch(:unbookable) { Array.new }).to_set
+  def initialize(opts = {})
+    self.attributes = attributes.merge(opts.symbolize_keys)
```
threedaymonk: Do you need this line? This is what the `initialize` method in Virtus does, as I understand it.

app/models/prison.rb
```diff
@@ -1,52 +1,88 @@
...
   end
 
   def anomalous_dates
-    (fetch(:slot_anomalies) { Hash.new }).keys.to_set
+    (slot_anomalies || {}).keys.to_set
   end
 
   def days_lead_time
-    fetch(:lead_days, DEFAULT_LEAD_DAYS)
+    lead_days
```
threedaymonk: Can we rationalise these two methods to one?

app/models/prisoner.rb
```diff
@@ -10,14 +10,21 @@ class Prisoner
   }
 
   validates :prison_name, inclusion: {
-    in: Rails.configuration.prison_data.keys
-  }
+    in: Prison.names,
+    message: "must be chosen" }
```
threedaymonk: Please put new error messages in `config/locales/en.yml`.

app/views/prisoner_details/edit.html.haml
```diff
@@ -30,42 +29,49 @@
       %fieldset
         %legend.visuallyhidden Prisoner details
 
-        =group_container(f, :first_name) do
+        .group{ :class => ('validation-error' if f.object.errors.include?(:first_name)) }
```
threedaymonk: Check your merge here: you're destroying my earlier work.

tatyree: Sorry.  I'll review. I'm finding it challenging to keep up with the changes.

config/initializers/prisons.rb
threedaymonk: 極度乾燥

app/views/prisoner_details/edit.html.haml
```diff
@@ -30,42 +29,49 @@
...
+        .group{ :class => ('validation-error' if f.object.errors.include?(:prison_name)) }
           = f.label :prison_name do
             Name of the prison
             %p.form-hint eg Cardiff
-            =field_error(f, :prison_name)
+            %span.validation-message= f.object.errors[:prison_name].first if f.object.errors.include?(:prison_name)
           - if f.object.errors.include?(:prison_name_reason)
             %p
-              - case prison_data['reason']
+              - case prison.try(:reason)
```
threedaymonk: Why do we need `try` here? is `prison` ever nil?

threedaymonk: Incidentally, I think this line should be the only change in this file.

tatyree: I believe it was in some of the specs when I first came across that error.  I believe that is fixed now.  I'll change it over and test. 

As regards your "only change in this file" comment - you mean as regards the merge error killing your view work?

app/models/prisoner.rb
tatyree: Will do

app/models/prison.rb
tatyree: It wasn't overriding the defaults correctly unless I did the merge manually.  I'll take another look now that I've got a bit of distance from it, but this was the only was I could keep it from always preferring the default.

app/helpers/visit_helper.rb
tatyree: Possibly not.  However, it isn't something we're testing for when we look at the source data.

app/views/prisoner_details/edit.html.haml
threedaymonk: Yes: I think all the others are merge errors.

app/views/confirmations/new.html.haml
threedaymonk: For our next trick, perhaps we can rename `canned_responses` to something more descriptive: it doesn't seem awfully clear to me.

app/views/prisoner_details/edit.html.haml
tatyree: It definitely needs the try, otherwise it raise a NoMethodError when a user clicks on continue without first selecting a prison: 
spec/features/prison_information_page_spec.rb:10

app/models/prison.rb
dan-palmer: Bump. Quite sure we can rely on `Virtus` `initialize` implementation.

tatyree: I'll check it again.  It wasn't working correctly when I tried it.  Issue with the specs, maybe.

app/models/prison.rb
```diff
@@ -1,52 +1,86 @@
...
+    new(opts).tap { |p| Rails.configuration.prisons << p }
   end
 
-  attr_reader :name
+  def self.enabled
+    all.select(&:enabled)
+  end
 
-  delegate :fetch, to: :@prison_hash
+  def self.find(name_or_nomis)
```
dan-palmer: This change eliminates the requirement for `PrisonNotFound` error. Is this intentional?

tatyree: Yep.  The only place it was still being called was in the spec.

dan-palmer:
I don't think we should go with the autoloading all prison objects on startup approach due to the objects being mutable. It's a potential source of a class of bugs we can just straight up avoid. Instead I propose that we extract the attributes from the YAML file on demand and inject them into the constructor when required. Any change to a prison object in its lifetime would now be scoped to just that, its lifetime.

tatyree:
@dan-palmer I'm happy with that.  However, I've already been delayed multiple times by scope-creep.  Let me get it released as-is and then raise a ticket to change this behaviour.  Otherwise, I'm never going to get it out the door.

threedaymonk:
@tatyree Unfortunately, I think @dan-palmer raises an important point. Or, fortunately, that he's raised it *before* we released this. I don't have confidence in the codebase keeping the underlying state consistent across requests.

tatyree:
@threedaymonk on the one hand I'd argue that we're no worse off than we were with the hash data structure, which did exactly the same thing.  On the other...yeah...you're both right :sob:...I'll do it now...

app/models/prison.rb
dan-palmer: What do you think about this approach:

```ruby
def self.find(identifier, attribute: :name)
  all.find { |p| p.public_send(attribute) == identifier }
end
```

tatyree: :+1:

threedaymonk: My concern here is that this looks a lot like ActiveRecord's interface, but not completely: it's in the uncanny valley a bit. Perhaps we could go for the simple:

```ruby
def self.find_by_name(name)
  all.find { |p| p.name == name }
end
```

and do the same for `find_by_nomis_id`.

dan-palmer: I think that suggestion works well. My concern was with the or in the arg 👍

README.md
```diff
@@ -23,11 +21,17 @@ All known prisons should exist in the data files. If a prison is not in scope of
...
+- nomis_id: RCI
+  enabled: true
+  name: Rochester
+  ...
+
+- nomis_id: SFI
   enabled: false # this prison does not accept online visit request through this service
+  name: Stafford
+  ...
+
```
threedaymonk: Do these instructions also need updating?

app/models/prison.rb
```diff
@@ -1,52 +1,84 @@
...
 
-  delegate :fetch, to: :@prison_hash
+  def self.find(name)
+    all.detect { |p| p.name == name }.try(:dup)
+  end
 
-  def unbookable_dates
-    (fetch(:unbookable) { Array.new }).to_set
+  def self.find_by_nomis_id(nomis_id)
+    all.detect { |p| p.nomis_id == nomis_id }.try(:dup)
```
threedaymonk: Now that we've discovered that NOMIS IDs aren't unique per prison (or per our model of a "prison", anyway), is this method ever safe to use?

SteveMarshall:
:bomb: … :boom: :fireworks:

Amazing work getting this through, @tatyree! :hand:

PR #358, opened 2015-10-13
Fix typo in smoke tests

threedaymonk:
This didn't break anything, as the misspelling was consistent, but it was a bit disconcerting.

### DIFF
--- a/smoke_test/smoke_test.rb
+++ b/smoke_test/smoke_test.rb
@@ -17,10 +17,10 @@ require_relative 'steps/prison_booking_confirmation_copy'
 require_relative 'steps/cancel_booking_page'
 require_relative 'steps/prison_booking_cancelled'
 
-module SuprressJsConsoleLogging; end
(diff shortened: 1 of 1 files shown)


dan-palmer:
:+1:

smoke_test/smoke_test.rb
```diff
@@ -17,10 +17,10 @@
 require_relative 'steps/cancel_booking_page'
 require_relative 'steps/prison_booking_cancelled'
 
-module SuprressJsConsoleLogging; end
```
dan-palmer: :sob:

EGiataganas:
@threedaymonk :+1:

smoke_test/smoke_test.rb
dan-palmer: Might this be a little nicer on the eyes as:
```ruby
SuppressJsConsoleLogging = Module.new
```

PR #359, opened 2015-10-13
Change date format

EGiataganas:
Date format changed for both 'prison_mailer' and 'visitor_mailer' to
be (full weekday name, day of the month, full month name and year).
So both of them have the same date format and give all the necessary
details


threedaymonk:
Good work. It's always nice when things get more consistent.

For clarification: @EGiataganas checked with @tonyduarte about these formats. They weren't supposed to be different, and they should have included the day and year.

SteveMarshall:
:stars:

PR #362, opened 2015-10-13
Record an event when an input fails validation

threedaymonk:
We don't have any visibility of what happens when a visitor's browser prevents them from submitting a form because an input failed validation, so we don't know when we have inadvertently locked them out.

The HTML5 form validation API exposes an 'invalid' event that we can listen for. We can then POST back to the server with some salient details (path and field name) so that we can monitor and graph these over time, and be aware of spikes that indicate when we've messed up.

Logstasher will record the parameters, and this can be viewed in Kibana with a query like:

    (type:pvb AND path:"/frontend_events")

Unfortunately, PhantomJS does not respond to the 'invalid' event, so we can't actually test this except by manual means. Sad times.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -22,6 +22,7 @@
 //= require modules/moj.Conditional
 //= require modules/moj.checkbox-summary
 //= require modules/moj.AgeLabel.js
+//= require modules/moj.form-errors
 
 // Assets without a better place
 //= require shame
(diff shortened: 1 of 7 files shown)


threedaymonk:
Tested on preprod: events are recorded and can be seen in Kibana.

EGiataganas:
@threedaymonk Nice!! :+1:

threedaymonk:
For the attention of @ministryofjustice/frontend.

maxf:
looks good to me

PR #365, opened 2015-10-14
Rename METRICS_AUTH_KEY_NAME

EGiataganas:
We rename ENV['METRICS_AUTH_KEY'] to ENV['TRUSTED_USERS_ACCESS_KEY']
so we can have better idea what it is for.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,7 +9,7 @@ class ApplicationController < ActionController::Base
   end
 
   def reject_without_key_or_trusted_ip!
-    unless valid_auth_key? || permitted_ip?
+    unless valid_trusted_users_access_key? || permitted_ip?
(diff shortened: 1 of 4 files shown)


threedaymonk:
Is there a corresponding change to the server configuration?

app/controllers/application_controller.rb
```diff
@@ -20,7 +20,7 @@ def permitted_ip?
   end
 
   def auth_key
```
SteveMarshall: Probably worth renaming this, too?

SteveMarshall:
@threedaymonk Yeah. I've added the new key to the server configuration (cc @EGiataganas) alongside the old one, so both will work for now. Once we've merged this, I'll remove the old one.

EGiataganas:
@SteveMarshall @threedaymonk Tested.

PR #366, opened 2015-10-15
Tidy lib directory

threedaymonk:
The lib directory is a disorganised junk drawer. This moves the files found there to more appropriate locations.


spec/services/email_checker_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'rails_helper'
-require 'email_validator'
+require 'email_checker'
```
dan-palmer: I think this `require` is redundant

lib/tasks/static_pages.rake
```diff
@@ -20,7 +20,7 @@ namespace :static_pages do
       outpath = File.join([Rails.root, 'public', output])
       resp = app.get(route)
       if resp == 200
-        File.delete(outpath) unless not File.exist?(outpath)
+        File.delete outpath if File.exist?(outpath)
```
dan-palmer: still :laughing:

spec/models/prison_day_spec.rb
```diff
@@ -81,12 +81,16 @@
...
 
         context 'and has an anomalous booking slot' do
-          subject { described_class.new(
-            bank_holiday_friday,
-            prison_from(prison_data.merge(
-              slot_anomalies: {bank_holiday_friday => ["0930-1000"]}
-            ))
-          )}
+          subject {
+            described_class.new(
```
dan-palmer: :+1: :pray:

spec/services/email_checker_spec.rb
threedaymonk: I think you're right.

lib/tasks/static_pages.rake
SteveMarshall: I don't not like undoing the addition of `unless not` :+1:

threedaymonk: I could not fail to disagree with both of you less.

threedaymonk:
This is now waiting on #357, which is likely to precipitate some merge conflicts.

lib/tasks/static_pages.rake
tatyree: All I can not succeed at deducting from this conversation is "Buffalo buffalo Buffalo buffalo buffalo buffalo Buffalo buffalo"

threedaymonk: John, where Mark had had had had had had had. Had had had had a better effect on the teacher.

PR #367, opened 2015-10-16
Add logging for legacy prison names

threedaymonk:
We believe that these are no longer needed, but in order to be more confident we'll put a message into the simple Rails log because we know we can track it there.


threedaymonk:
On investigating Grafana, we've determined that these are no longer used.

PR #368, opened 2015-10-16
Remove legacy data fixes

threedaymonk:
These prison names were last changed in 2014, and we've checked the metrics and verified that these fixes are no longer used.

Whilst we may need to handle prison name changes in the future, we will do this in a better way than by dumping a hash of strings directly into the controller.

### DIFF
--- a/app/controllers/confirmations_controller.rb
+++ b/app/controllers/confirmations_controller.rb
@@ -67,8 +67,7 @@ class ConfirmationsController < ApplicationController
   end
 
   def booked_visit
-    @booked_visit ||=
-      legacy_data_fixes(encryptor.decrypt_and_verify(params[:state]))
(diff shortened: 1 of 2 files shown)


app/controllers/confirmations_controller.rb
```diff
@@ -88,25 +87,6 @@ def confirmation_params
...
-  }
-
-  def legacy_data_fixes(visit)
-    if LEGACY_PRISON_NAMES.key?(visit.prison_name)
-      STATSD_CLIENT.increment('pvb.app.legacy_data_fixes')
-      visit.prison_name = LEGACY_PRISON_NAMES[visit.prison_name]
-    end
-    visit
-  end
-
```
tatyree: :+1:

dan-palmer:
:pray:

PR #370, opened 2015-10-19
Fix first_slot_date_visitor_format method

EGiataganas:
We refactored first_slot_date_visitor_format to use locale file.


dan-palmer:
Will this work? The smoke tests aren't run via Rails so `I18n` won't be loaded.

dan-palmer:
Shouldn't we follow a sort double booking pattern here and have the date format duplicated for the smoke tests. I feel as though this directory is in the repo out of convenience and shouldn't draw on any of the assets.

EGiataganas:
@dan-palmer you re right.

PR #372, opened 2015-10-19
Remove date_of_birth_native and tidy controllers

threedaymonk:
`date_of_birth_native` did not exist in the interface, so there's no need to handle it on the backend.

In the process of doing this, I noticed some simple improvements that I could make to simplify date handling and to make these two controllers more straightforward.

We now use Virtus's date handling, which is simpler to grasp and obviates the need to translate to and from Rails's peculiar format (`1i`/`2i`/`3i` etc.).


app/controllers/prisoner_details_controller.rb
```diff
@@ -29,22 +30,10 @@ def update
...
-    end
-    trim_whitespace_from_params %i[
-      first_name last_name date_of_birth number prison_name
-    ]
-  rescue ArgumentError
-    trim_whitespace_from_params %i[first_name last_name number prison_name]
+    params.require(:prisoner).permit(
+      :first_name, :last_name, :number, :prison_name,
+      date_of_birth: [:day, :month, :year]
+    )
```
EGiataganas: @threedaymonk great improvement!!

app/controllers/visitors_details_controller.rb
```diff
@@ -1,22 +1,54 @@
...
 
   def edit
+    visit.visitors = [Visitor.new] if visit.visitors.empty?
     @collect_phone_number = true
   end
 
-  def this_path
-    edit_visitors_details_path
+  def update
+    m = params[:next].match(/remove-(\d)/)
```
dan-palmer: I think lines 14-21 need a little investigation at some point. I reckon `m` is always `nil`. 😓

dan-palmer: `build_visitors` will always just replace the current visitors in the session

SteveMarshall: I think it is always nil… Unless you add a second visitor, I think? (I think this is matching the “remove visitor from booking” buttons.)

threedaymonk: This chunk is used when you remove a visitor, as @SteveMarshall says. (At least, if you delete it, that's the part of the specs that fails.)

dan-palmer: I can't seem to see anything in the repo that sets the `next` param to the remove + visitor index for this to work which makes me think you always just have `next` being equal to 'Complete' and because the JS has removed the fields from the DOM when the remove link is clicked the code just works by rebuilding the visitors again in the backend from the that requests params. I hope I'm just being stupid here and missing the obvious.

dan-palmer: Ooo is this for JS being turned off perhaps?

threedaymonk: Ah yes, I see what you mean. There's one way that uses `next; 'remove-1'` etc., which doesn't seem to be exercised anywhere except in one spec (and is thus presumably just junk code), and another that sets a `_destroy` attribute – that presumably doesn't work without JavaScript.

dan-palmer: Answer to above: nope its not for JS being unavailable.

Manually tested this and it uses the `_destroy` field to check if it should be built or not.

threedaymonk:
And it uses the `_destroy` field with JS turned off, too. So that `next` bit *can* be deleted.

app/controllers/visitors_details_controller.rb
dan-palmer: Yeah assumed the tests are explicitly passing something to exercise this otherwise irrelevant code. 😔

threedaymonk: And it uses the `_destroy` field with JS turned off, too. So that `next` bit *can* be deleted.

PR #373, opened 2015-10-19
Remove smtp_alive? (again!)

threedaymonk:
This was deleted in 9c276713 but was inadvertently reintroduced in ac7605bd.


SteveMarshall:
So I might actually argue that we should re-add the use of this to the health check: it's a useful indicator, alongside the queue sizes, of overall health.

tatyree:
@SteveMarshall I went ahead and merged this and added a ticket to the icebox to reintroduce it later: 

https://www.pivotaltracker.com/story/show/106093170

SteveMarshall:
@tatyree :+1:

PR #374, opened 2015-10-19
Randomly pad executions that overrun

SteveMarshall:
We've had a couple of instances where the padding time wasn't sufficient for the execution time of the block. This causes `sleep` to raise an exception because of a negative sleep time.

This logs when we are insufficiently padding, and them always applies a random padding based on the execution time to make it harder to carry out timing attacks.


SteveMarshall:
@tatyree This is what I came up with to avoid the 5XX error in `pad_execution_time` you found. Waddya think?

app/controllers/application_controller.rb
```diff
@@ -77,11 +77,17 @@ def ensure_visit_integrity
...
-  def pad_execution_time(execution_time)
+  def pad_execution_time(execution_time_limit, max_pad_multiple=5)
     start = Time.zone.now
     result = yield
     stop = Time.zone.now
-    sleep execution_time - (stop - start)
+    pad_time = execution_time_limit - (stop - start)
+    if pad_time < 0
+      pad_time = rand(max_pad_multiple) * execution_time_limit
+      Rails.logger.info "Execution padding insufficient"
```
SteveMarshall: @threedaymonk What do we get out of Kibana with this? (I based it on some logging you did.) Do we get the call stack so we can see the origin, or should I do some faffing here to pass that through to the logs?

threedaymonk: It comes out like this:

```json
{
  "_index": "logstash-2015.10.13",
  "_type": "pvb",
  "_id": "2d0f34af-1a3d-3aed-6a72-b6e5e04960b9",
  "_score": 4.2507687,
  "_source": {
    "tags": [
      "ruby",
      "application",
      "production",
      "pvb"
    ],
    "@version": 1,
    "@timestamp": "2015-10-13T21:37:12.327+00:00",
    "host": "front-01.prod1.pvb",
    "file": "/srv/pvb/application/shared/log/production.log",
    "message": "I, [2015-10-13T21:37:11.224153 #13003]  INFO -- : Deserializing legacy token",
    "type": "pvb",
    "@logstash_uuid": "2d0f34af-1a3d-3aed-6a72-b6e5e04960b9"
  }
}
```

In the case of the legacy tokens, I just used the literal text to select entries for the graph.

SteveMarshall: Hm. That's annoying; it'd be nice to have the stack trace—or at least the caller—so we can identify which calls are overrunning. Any ideas?

threedaymonk:
It occurs to me that this might actually *not* be a safe solution: repeating the same input would allow an attacker to eliminate the random component.

SteveMarshall:
@threedaymonk A good point.

@ohmygauss: do you have any thoughts on how we might safely extend the time of an intended-to-be-constant execution when it overruns?

app/controllers/application_controller.rb
threedaymonk: I don't know how easy it is to add arbitrary fields to logstash (my initial research said not very), but you can get the call stack with `caller` (it's an array of strings, so something like `caller.join("\n")` might be more useful).

tatyree:
@SteveMarshall once we've heard back from @ohmygauss, could we get a couple of specs on this, please?

SteveMarshall:
@tatyree Yeah, definitely. I might end up wanting to include `Minitest::Benchmark` so I can use [`assert_performance_constant `](https://github.com/seattlerb/minitest/blob/master/lib/minitest/benchmark.rb#L105…L136) to test the timeliness of this, I think.

tatyree:
@SteveMarshall sounds like a plan.  At a minimum, I'd suggest something like this should pass: 

```
  describe '.pad_execution_time' do
     it 'does not raise exception if padding time is lower than run time' do
        expect{
          controller.pad_execution_time 0.1 do
            sleep 0.2
          end
        }.not_to raise_exception
     end
  end
```

SteveMarshall:
@tatyree :+1:

ohmygauss:
So pad time shouldn't be a random multiple of max_pad, but a fixed one. Through some load testing I'd imagine you would have a better idea of the worst case scenario. The pad time should be a fixed (smallest possible) multiple to account for the worst case scenario.

Ideally the intended padding time should account for exceptions.

tatyree:
@SteveMarshall based on @ohmygauss's comment, I vote we put in a test, a bit of logging and raise the padding on the current calls to 0.2 or 0.3 and see what happens.

dan-palmer:
If we are concerned about timing attacks and leaking key lengths why can't we just hash the inputs first (e.g. bcrypt) and then use `ActiveSupport::SecurityUtils.secure_compare` ?

ohmygauss:
Do you mean some kind of HMAC and then secure_compare?

SteveMarshall:
@dan-palmer The [argument against `secure_compare`](https://github.com/ministryofjustice/prison-visits/pull/338#discussion_r41008434) is that we'd have to HMAC both sides (or something) before passing through `secure_compare` to avoid a timing attack.

SteveMarshall:
@dan-palmer Oh, that's what you said :disappointed_relieved:

SteveMarshall:
It seems we've concluded that padding the execution time is the wrong way to do this and, instead, we should generate HMACs for the inputs as I think we discussed when I was originally working on this. :sweat:

PR #376, opened 2015-10-20
Amend Nottingham slot times by +15mins except Sat/Sun afternoon

kyliemulholland:


kyliemulholland:
Tested in preprod - passed.

PR #377, opened 2015-10-20
Prevent emails being sent outside production

tatyree:
It should not be sending emails to the prisons from preprod, test or
dev.  I missed this when merging prison_data_production and
prison_data_staging.  This fixes it programatically and for all prisons.

### DIFF
--- a/config/initializers/prisons.rb
+++ b/config/initializers/prisons.rb
@@ -1,5 +1,9 @@
 Rails.configuration.prisons = []
 
 Dir["config/prisons/*.yml"].each { |file|
-  Prison.create(YAML.load_file(file))
+  prison = YAML.load_file(file)
+  unless Rails.env.production?
+    prison['email'] = "pvb.#{prison['nomis_id']}@maildrop.dsd.io"
+  end
+  Prison.create(prison)
 }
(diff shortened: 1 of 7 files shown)


config/initializers/prisons.rb
```diff
@@ -1,5 +1,9 @@
 Rails.configuration.prisons = []
 
 Dir["config/prisons/*.yml"].each { |file|
-  Prison.create(YAML.load_file(file))
+  prison = YAML.load_file(file)
+  unless Rails.env.production?
+    prison['email'] = "pvb.#{prison['nomis_id']}@maildrop.dsd.io"
```
SteveMarshall: I'm not sure this is correct, is it? For example, it's `pvb.cardiff@maildrop.dsd.io`, if memory serves…

tatyree: It was line of least resistance.  Using names means having to parameterize...Can change it once I've got a Site association in place.  This was just to unblock @threedaymonk on testing on preprod.

tatyree: And, also, not all of the prison_data_staging prisons from the old format had maildrop.dsd.io emails...

SteveMarshall: Fair enough! Go ahead then. As for not having maildrop emails everywhere… :sob:

PR #378, opened 2015-10-20
Do not show disabled prisons in metrics

tatyree:
The metrics_controller was using Prison.nomis_ids, which exposed the
nomis_ids for all prisons, not simply the enabled ones. I believe this
was causing some cruft to appear in /metrics.  This should fix that.

### DIFF
--- a/app/controllers/metrics_controller.rb
+++ b/app/controllers/metrics_controller.rb
@@ -1,8 +1,8 @@
 class MetricsController < ApplicationController
   permit_only_trusted_users
+  before_action :nomis_ids, only: [:index, :weekly]
 
   def index
-    @nomis_ids = Prison.nomis_ids
     if params[:range] == 'all'
       @dataset = CalculatedMetrics.new(VisitMetricsEntry.deferred, 3.days)
     else
@@ -66,7 +66,6 @@ class MetricsController < ApplicationController
       @start_of_year,
       ApplicationHelper.instance_method(:prison_estate_name_for_id)).
     refresh
-    @nomis_ids = Prison.nomis_ids
 
     respond_to do |format|
       format.html
@@ -74,6 +73,10 @@ class MetricsController < ApplicationController
         render text: @dataset.csv
       end
     end
+  end
+
+  def nomis_ids
+    @nomis_ids = Prison.enabled_nomis_ids
   end
 
   def fortnightly_range
--- a/spec/models/prison_spec.rb
+++ b/spec/models/prison_spec.rb
@@ -51,179 +51,182 @@ RSpec.describe Prison, type: :model do
         end
       end
 
-      describe '.enabled' do
-        it 'returns only enabled prisons' do
-          expect(described_class.enabled.all?(&:enabled?)).to be_truthy
+      context 'collections' do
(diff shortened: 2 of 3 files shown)


spec/models/prison_spec.rb
```diff
@@ -91,6 +91,18 @@
         expect(described_class.nomis_ids).to match_array(['WSI'])
       end
     end
+
+    describe '.enabled_nomis_ids' do
+      before do
+        allow(described_class).to receive(:all).
+          and_return([double('prison', nomis_id: 'WSI').as_null_object,
```
dan-palmer: Suggestion: `object_double`, in order to ensure the public interface to the model is correct.

```ruby
object_double(described_class.new, nomis_id: 'WSI').as_null_object
```

app/controllers/metrics_controller.rb
```diff
@@ -76,6 +75,10 @@ def weekly
     end
   end
 
+  def nomis_ids
+    @nomis_id = Prison.enabled_nomis_ids
```
threedaymonk: Shouldn't this be `@nomis_ids` (plural)?

spec/models/prison_spec.rb
```diff
@@ -84,13 +84,25 @@
...
+                      object_double(described_class.new, nomis_id: nil).as_null_object])
       end
       it 'returns nomis ids for all prisons with a nomis id' do
         expect(described_class.nomis_ids).to match_array(['WSI'])
       end
     end
+
+    describe '.enabled_nomis_ids' do
+      before do
+        allow(described_class).to receive(:all).
```
dan-palmer: I think it might be better to stub the Rails config here as opposed to this specific class method -`.all`. We are revealing knowledge of the implementation here where as I feel stubbing the `data store` doesn't do this as much as its the source of truth for the models. We are saying that we know that `enabled` must call `all` before doing its filtering.

spec/models/prison_spec.rb
```diff
@@ -51,179 +51,182 @@
...
-      describe '.enabled' do
-        it 'returns only enabled prisons' do
-          expect(described_class.enabled.all?(&:enabled?)).to be_truthy
+      context 'collections' do
+        let(:enabled)  { object_double(described_class.new, nomis_id: 'ENA', name: 'Enabled', enabled: true) }
+        let(:disabled) { object_double(described_class.new, nomis_id: 'DIS', name: 'Disabled', enabled: false) }
+        let(:no_nomis) { object_double(described_class.new, nomis_id: nil, name: 'No Nomis', enabled: false) }
+
+        let(:prisons) {[enabled, disabled, no_nomis]}
+        before do; allow(Rails.configuration).to receive(:prisons).and_return(prisons) end
```
dan-palmer:
```ruby
before { allow(Rails.configuration).to receive(:prisons).and_return(prisons) }
```

spec/models/prison_spec.rb
```diff
@@ -51,179 +51,182 @@
         end
       end
 
-      describe '.enabled' do
-        it 'returns only enabled prisons' do
-          expect(described_class.enabled.all?(&:enabled?)).to be_truthy
+      context 'collections' do
+        let(:enabled)  { object_double(described_class.new, nomis_id: 'ENA', name: 'Enabled', enabled: true) }
```
dan-palmer: This is much better! :+1:

PR #379, opened 2015-10-20
Remove open responses

threedaymonk:
Since 5cf680f4, all\* prisons have used only canned responses. Open responses are not used, and can be removed.

\* Well, almost. They were *supposed* to, but Foston Hall was omitted from that commit. So this also fixes that problem.


dan-palmer:
:boom: :+1:

EGiataganas:
:sunglasses: :rocket:

PR #380, opened 2015-10-20
Record NOMIS ID in Kibana

threedaymonk:
At present, we record the prison name, but those are sometimes multiple per NOMIS ID.


SteveMarshall:
@threedaymonk Do we still want to do this? If so, can you rebase against master, and I'll merge…

threedaymonk:
I've rebased.

PR #381, opened 2015-10-21
Add estate to every prison

tatyree:
The nomis_id is not unique amoung the different organisational units
that comprise some prisons (Winchester Remand Only and Winchester
Convicted Only, for example). However, it is used for calculating
metrics. This causes errors as it unclear which unit the metric is
related to.  Adding an organisational unit of 'estate' should resolve
this as it will allow the report to show the higher-level organisation
the nomis_id actually appies to ('Winchester' here, for instance).

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -55,9 +55,9 @@ module ApplicationHelper
     end
   end
 
-  def prison_name_for_id(nomis_id)
+  def prison_estate_name_for_id(nomis_id)
     prison = Prison.find_by_nomis_id(nomis_id)
-    return prison.name if prison && prison.enabled?
+    return prison.estate if prison && prison.enabled?
   end
 
   def markdown(source)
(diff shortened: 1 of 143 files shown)


app/helpers/application_helper.rb
```diff
@@ -55,9 +55,11 @@ def conditional_text(variable, prefix = '', suffix = '')
     end
   end
 
+  # This was returing the name, but now returns the estate name -
+  # the top-level organising unit for a particular prison
   def prison_name_for_id(nomis_id)
```
dan-palmer: Rather than have a comment here shouldn't we alter the method name to reveal the thing its returning?

tatyree: @dan-palmer normally I would have, but it's only used in metrics which we're going to refactor soon.  Wanted to keep the effort on this to a minimum.

dan-palmer:
:star2: :+1: Do think changing the name was worthwhile.

tatyree:
I took a second look and realised it wasn't as odious as I thought it was going to be...

tatyree:
I still hate metrics, though :angry:

PR #382, opened 2015-10-22
Use Prison to update to nomis_ids in metrics table

tatyree:
This should have been part of the overall change to the Prison model,
but got missed for some reason.


db/migrate/20150401151513_use_nomis_ids_for_metrics.rb
```diff
@@ -1,8 +1,8 @@
 class UseNomisIdsForMetrics < ActiveRecord::Migration
   def change
-    Rails.configuration.prison_data.each_pair do |prison_name, data|
-      say_with_time "Renaming #{prison_name} to #{data['nomis_id']}" do
-        VisitMetricsEntry.where(prison_name: prison_name).update_all(prison_name: data['nomis_id'])
+    Prison.all.each do |prison|
+      say_with_time "Renaming #{prison.name} to #{prison.nomis_id}"do
+        VisitMetricsEntry.where(prison_name: prison.name).update_all(prison_name: prison.nomis_id)
```
threedaymonk: Don't use ActiveRecord classes in migrations: if the class definition changes, the migration becomes impossible to run.

There are two alternatives:

1. Copy the AR model inside the migration, freezing it at the desired version.
2. Use SQL.

PR #383, opened 2015-10-22
Send to maildrop in test or based on env

tatyree:
This was being selected for everything but production, but preprod, of
course, runs in production mode, so it wasn't working. This fixes that.


EGiataganas:
Tested on preprod and is working as it should be!! :+1:

PR #384, opened 2015-10-23
Remove model access from migrations

tatyree:
These are no longer need and all relied on the `VisitMetricsEntry` model, which made them fragile. Several of the earlier ones operated on the `prison_name` attribute, which a later one changes to be the `nomis_id` attribute (why not just *add* `nomis_id`, instead of renaming `prison_name`?).

A couple were also using VisitMetricsEntry to effectively set defaults for columns post-hoc.


app/models/visit_metrics_entry.rb
```diff
@@ -2,7 +2,7 @@ class VisitMetricsEntry < ActiveRecord::Base
   validates_presence_of :visit_id, :nomis_id, :requested_at
   validates_inclusion_of :kind, in: %w{deferred}
```
dan-palmer: Since `instant` has been shown the :door:  why do we still have `kind`?

tatyree: @dan-palmer because I didn't realise what it was for when I was shuffling migrations...

dan-palmer:
Regarding the default values being set in the migrations I think we should have them declared in the model instead. We currently do something similar in the `Prison` model.

tatyree:
I'm uncomfortable with using the model to set defaults on persisted models.  Spent too much time chasing down strange race conditions when someone's come along and done strange things to the code later.

SteveMarshall:
@dan-palmer What's the advantage of setting them in the model rather than the DB? I'd probably tend to doing it in the DB, as @tatyree is here, but maybe there's something we're missing?

dan-palmer:
Some of the models(non persisted though) have defaults declared in them through virtus. I'm not against @tatyree's approach - was just an observation around consistency really.

SteveMarshall:
So I reckon we leave this as-is. @tatyree, gimme a shout when you're ready for this to go.

PR #385, opened 2015-10-26
Move validity check to model and log missing info

tatyree:
We are experiencing 500 errors on the live service that *appear* to be
happening when existing visits are deserialised. Visits lose their
prisoner_name and prison.slots attributes.  This commit removes the
bulk of the Visit integrity checking logic to the model and uses
that method, `required_information?`, to call a series of logging methods
that look for missing prisoner and prison associations.  Hopefully, this
will give us some insight on where these errors are coming from.

### DIFF
--- a/app/models/visit.rb
+++ b/app/models/visit.rb
@@ -40,4 +40,24 @@ class Visit
   def same_visit?(other)
     other.visit_id == self.visit_id
   end
+
+  def prisoner?
+    prisoner.present?
+  end
+
+  def prisoner_number?
+    prisoner? && prisoner.number.present?
+  end
+
+  def prison_name?
+    prisoner? && prisoner.prison_name.present?
+  end
+
+  def prison?
+    prisoner && prison.present?
+  end
+
+  def prison_slots?
+    prison? && prison.slots.present?
+  end
 end
--- a/app/services/visit/integrity_checker.rb
+++ b/app/services/visit/integrity_checker.rb
@@ -1,0 +1,58 @@
+class Visit
+  module IntegrityChecker
+    def ensure_visit_integrity
+      return if required_information?
+      log_any_missing_information
+      flash[:notice] = I18n.t(
+        :ensure_visit_integrity,
+        scope: 'controllers.shared'
+      )
+      redirect_to edit_prisoner_details_path
+    end
+
+    def required_information?
+      visit && visit.prisoner? && visit.prison_name?
+    end
+
+    def log_any_missing_information
+      prefix = "Visit #{visit.visit_id} is missing a"
+      log_missing_prisoner(prefix)
+      log_missing_prisoner_number(prefix)
+      log_missing_prison_name(prefix)
+      log_missing_prison(prefix)
+      log_missing_prison_slots(prefix)
+    end
+
+    private
+
+    def log_missing_prisoner(prefix)
+      unless visit.prisoner?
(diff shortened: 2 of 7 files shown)


app/models/visit.rb
```diff
@@ -40,4 +47,41 @@ def adult?(visitor)
   def same_visit?(other)
     other.visit_id == self.visit_id
   end
+
+  def log_any_missing_information
+    prefix = "Visit #{visit_id} is missing a"
```
dan-palmer: We can tidy the logging implementation up a bit here. We are consistently asking `prisoner` if its `blank?` - we should try and consolidate those calls down to one.

Also think we should extract all of this logging stuff out into an `Visit::IntegrityChecker`(or similar) object that can be used in the `ApplicationController`.

app/models/visit.rb
```diff
@@ -40,4 +47,41 @@ def adult?(visitor)
...
+  end
+
+  def log_missing_prison(prefix)
+    return if prisoner.blank? || prison.present?
+    Rails.logger.info("#{prefix} prison")
+  end
+
+  def log_missing_prison_slots(prefix)
+    return if prisoner.blank? || prison.blank?
+    return unless prison.slots.blank?
```
dan-palmer: This would only ever happen if the yaml file with the prison data in was merged with slot data removed for one or more prisons?

app/services/visit/integrity_checker.rb
```diff
@@ -0,0 +1,56 @@
...
+        scope: 'controllers.shared'
+      )
+      redirect_to edit_prisoner_details_path
+    end
+
+    def required_information?
+      return false unless visit
+      return false unless visit.prisoner?
+      return false unless visit.prison_name?
+      true
```
SteveMarshall: Isn't this just `visit && visit.prisoner? && visit.prisoner_name?`? All the `false unless` makes my head hurt, though, so I may be missing some wrinkle.

app/services/visit/integrity_checker.rb
```diff
@@ -0,0 +1,56 @@
...
+    private
+
+    def log_missing_prisoner(prefix)
+      return if visit.prisoner?
+      Rails.logger.info("#{prefix} prisoner")
+    end
+
+    def log_missing_prisoner_number(prefix)
+      return if !visit.prisoner? || visit.prisoner_number?
+      Rails.logger.info("#{prefix} prisoner number")
```
SteveMarshall: As above, I think the inversion of logic here (and in all the siblings of this) is a bit confusing (but that could be because it's early and I'm pre-:coffee:). When does this log? When there's a prisoner but no prisoner number? It might be clearer to wrap the if around the logger, or to make it a conditional line?

SteveMarshall:
@tatyree Looks like the test failures here are on all the new logging tests in this PR? Looks like the problem is because of logic inversion weirdness, as I mentioned in https://github.com/ministryofjustice/prison-visits/pull/385#discussion_r43859127.

Maybe rewrite them something like this:

```ruby
def log_missing_prisoner_number(prefix)
  Rails.logger.info("#{prefix} prisoner number") if visit.prisoner? && !visit.prisoner_number?
end
```

… Or use a block for the if; I don't mind either way. Certainly, using `if visit.prisoner? && !other_thing` makes it clearer that you want there to exist a prisoner and *not* have the other thing, and so on.

SteveMarshall:
:boom:

PR #389, opened 2015-11-03
Update Springhill visit slots

kyliemulholland:
Add extra visits:
27/11, 4/12, 11/12, 16/12, 23/12, 30/12, 08/01, 15/01, 22/01, 29/01

Unbookable:
26/12, 27/12


kyliemulholland:
Tested on preprod as far as 27/11 - passed.

PR #408, opened 2015-11-06
Update Coldingly Christmas visit slots

kyliemulholland:
Unbookable - Christmas Day, Boxing Day, New Year's Day
Additional - 21, 22, 24, 28, 31 Dec 0930-1100
Slot anomaly - 31st Dec 1400-1530 instead of 1400-1545


EGiataganas:
@kyliemulholland Tested locally.

PR #422, opened 2015-11-11
Put metrics DB defaults into consistent state 

EGiataganas:


threedaymonk:
Can you add to the commit to say something like:

> Because an identical previous migration was deleted, the database on the CI server is in an inconsistent state. Duplicating the migration should ensure that all records have an `outcome`.

I think it might also be good to add a further migration to prevent a null `outcome`.

PR #430, opened 2015-11-19
Create HDI-hatfield-lakes

kyliemulholland:
Presuming it's ok for this site to have the same NOMIS ID as Hatfield Open, whilst we have no interface with NOMIS?


EGiataganas:
Tested locally.

PR #478, opened 2015-11-26
Update Grendon Christmas visit slots

kyliemulholland:
Unbookable:
- 18th December
- Christmas Day
- Boxing Day
- New Year's Day

Updated phone number


EGiataganas:
Tested locally

PR #486, opened 2015-11-27
Update Foston Hall 22nd December

kyliemulholland:
Make 22nd December unbookable


SteveMarshall:
The 22nd of December is already unbookable at Boston Hall because it's a Tuesday, so this change isn't needed.

PR #487, opened 2015-11-27
Add christmas updates

KamalaHB:
Adding christmas visits for highpoint north and south. 
Adding for christmas visits for buckley hall 
Adding Christmas visits for Stafford.


SteveMarshall:
Tested locally; LGTM! :tophat:

PR #495, opened 2015-11-30
Detect unbookable and anomalous date conflicts

tatyree:
We've been doing this manually, which is both time consuming and error
prone.  This spec automates the process.  It includes a working-out
of the acutal approach to aid in onboarding developers and support staff
in understanding the nature of the problem.


spec/configuration/prison_data_spec.rb
```diff
@@ -20,18 +20,48 @@
   context "enabled prisons" do
     let(:enabled) { Prison.enabled }
 
+    it "currently number between 95 and 105" do
+      expect(enabled.count).to be_within(5).of(100)
+    end
```
tatyree: I don't really like this approach, but I need to be sure that somebody doesn't accidentally unplug the loading of the live prison data.  If that happened, then some of the specs may well start to throw false positives.

spec/configuration/prison_data_spec.rb
```diff
@@ -20,18 +20,48 @@
...
     end
 
-    it "each unbookable date for each prison should be a valid date" do
-      unbookable = enabled.map(&:unbookable).flatten
-      expect(unbookable.map(&:class).all?{ |p| p == Date }).to be_truthy
-    end
+    context 'unbookable dates' do
+      it "are all valid dates" do
+        unbookable = enabled.flat_map(&:unbookable)
+        expect(unbookable.map(&:class).all?{ |p| p == Date }).to be_truthy
```
threedaymonk: We can use [the `all` matcher](https://relishapp.com/rspec/rspec-expectations/v/3-4/docs/built-in-matchers/all-matcher) here:

```ruby
expect(unbookable).to all(be_a(Date))
```

spec/configuration/prison_data_spec.rb
```diff
@@ -20,18 +20,48 @@
...
+      end
 
-    it "each unbookable date for each prison should be a unique" do
-      enabled.each do |prison|
-        expect(prison.unbookable.size).to eq(prison.unbookable.uniq.size)
+      it "are all unique" do
```
threedaymonk: It are tortuous grammar :stuck_out_tongue_winking_eye:

spec/configuration/prison_data_spec.rb
```diff
@@ -20,18 +20,48 @@
...
+          before do
+            allow(subject).to receive(:slot_anomalies).
+              and_return(Time.zone.today => [])
+            allow(subject).to receive(:unbookable).
+              and_return([Time.zone.today, Time.zone.tomorrow])
+          end
+
+          specify do
+            expect(subject.anomalous_dates - subject.unbookable_dates).
+              not_to eq(subject.anomalous_dates)
```
threedaymonk: I think logic operators might make more sense here:

```ruby
expect(subject.anomalous_dates & subject.unbookable_dates).to be_empty
```

(I assume there's no ready-made `expect(a).not_to intersect(b)` matcher, though that would be nice.)

spec/configuration/prison_data_spec.rb
tatyree: It are indeed.

spec/configuration/prison_data_spec.rb
tatyree: That would be.

PR #500, opened 2015-11-30
Change warning threshold for live prison count

tatyree:
I realised too late that number of HSE prisons due to be added over the
next couple of months was going to cause this to break.

### DIFF
--- a/spec/configuration/prison_data_spec.rb
+++ b/spec/configuration/prison_data_spec.rb
@@ -20,8 +20,8 @@ RSpec.describe "Prison data" do
   context "enabled prisons" do
     let(:enabled) { Prison.enabled }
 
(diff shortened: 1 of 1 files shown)


spec/configuration/prison_data_spec.rb
```diff
@@ -20,8 +20,8 @@
   context "enabled prisons" do
     let(:enabled) { Prison.enabled }
 
-    it "currently number between 95 and 105" do
-      expect(enabled.count).to be_within(5).of(100)
+    it "currently number at least 95" do
+      expect(enabled.count).to be >= 95
```
threedaymonk: I don't know what magic RSpec does internally to make this work, but it does work!

PR #527, opened 2016-01-05
Obviate the need for empty YAML defaults

threedaymonk:
By providing sensible defaults, we no longer need to put lines like `unbookable: []` in the YAML.

### DIFF
--- a/config/initializers/prisons.rb
+++ b/config/initializers/prisons.rb
@@ -1,7 +1,13 @@
 Rails.configuration.prisons = []
 
+default = {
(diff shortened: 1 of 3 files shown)


SteveMarshall:
Nice improvement! :bowtie: 

(But what about [the readme](https://github.com/ministryofjustice/prison-visits#non-bookable-days)? :wink:)

tatyree:
@SteveMarshall no one ever seems to read those, anyway :)

threedaymonk:
@SteveMarshall see #528

PR #529, opened 2016-01-12
Update Bullingdon remand visit slots

kyliemulholland:
Add Wednesday and Sunday slots
Update all visit times to 1345-1545


threedaymonk:
It's not essential, but it might be nice to put the days in order!

PR #531, opened 2016-01-18
Proxy certain paths to new app (prototype)

mloughran:
Prototypes the use of rack-proxy for this.

To test it out, provide the optional configuration variable which enables the proxying functionality.

		NEW_APP_URL=http://localhost:3000/ rails s -p 3001

This has uncovered a few issues:

* What to do about assets? Currently I've found yet another way to generate unstyled pages...
* What about HTTPS requests? Should they be proxied over SSL?

The paths to proxy currently hardcoded as /en/... and /cy/...

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,7 @@ gem 'curb'
 gem 'string_scrubber', '>= 0.2.0'
 gem 'virtus'
 gem 'sinatra'
+gem 'rack-proxy'
 
 group :test do
   gem 'codeclimate-test-reporter'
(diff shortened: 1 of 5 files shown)


mloughran:
To answer my question, `config.asset_host` is the obvious answer. Aren't we already uploading assets to S3?

threedaymonk:
This *doesn't* work when proxying to an HTTPS server. I can't immediately find the solution.

mloughran:
My conclusion is that we should use nginx for this if we can.

threedaymonk:
As we've settled on nginx for proxying, I'm going to close this.

PR #534, opened 2016-01-22
Update _standard_id_requirements.md

kyliemulholland:
@leighmoney I've added a new acceptable ID to the list - biometric residence permit. Are you happy with this content change?


threedaymonk:
I'll change this in the new app, too, when everyone is happy.

kyliemulholland:
Where would I find this on the new app @threedaymonk ? I was looking but couldn't figure out the path.

threedaymonk:
@kyliemulholland It's in the locale files, in [`config/locales/en/views.yml`](https://github.com/ministryofjustice/prison-visits-2/blob/master/config/locales/en/views.yml#L367).

